PR #12, opened 2017-02-08
Add machine ID as part of the hostname in /etc/hosts

grunskis:
Hostname initially is set correctly but `/etc/hosts` entry is missing the machine ID.


tomster:
i can see now, how this couldn't have worked otherwise :-)

i also like that we now define the full hostname only in one place.

PR #14, opened 2017-02-13
Tell HASS to stop installing requirements, do it manually

grunskis:
* Install homeassistant and let it install it's dependencies 
* Disable automatic pip intallation, preinstall Sonos / Philips Hue / HTTP libs

### DIFF
--- a/deployment/roles/hub-hass/templates/supervisor.conf
+++ b/deployment/roles/hub-hass/templates/supervisor.conf
@@ -1,9 +1,9 @@
 [program:nuimo_hass]
-command = {{deploy_location}}/venv/bin/hass -c {{data_location}}/
+command = {{deploy_location}}/venv/bin/hass -c {{data_location}}/ --skip-pip
 autostart=true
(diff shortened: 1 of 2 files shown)


deployment/roles/hub-hass/templates/supervisor.conf
```diff
@@ -1,9 +1,9 @@
 [program:nuimo_hass]
-command = {{deploy_location}}/venv/bin/hass -c {{data_location}}/
+command = {{deploy_location}}/venv/bin/hass -c {{data_location}}/ --skip-pip
 autostart=true
 autorestart=true
 directory={{ deploy_location }}
-stdout_logfile={{ data_location }}/wsgi.log
+stdout_logfile={{ data_location }}/hass.log
```
tomster: technically, this is not part of the PR

it's not a big deal in this specific context, so we can just merge it along, but in the future something like this should go into an extra commit, that is labelled something like "cleanup", "housekeeping" etc.

PR #19, opened 2017-02-22
Install Python3 from src

tomster:


larsblumberg:
LGTM. Only thing is, I would rename folder `python3-from-src` to `python3-from-source`

PR #21, opened 2017-02-28
Nuimo app role

grunskis:
Automate setup installation of Nuimo app.

To be used together with https://github.com/getsenic/senic-hub-nuimo-app/pull/1

### DIFF
--- a/deployment/roles/hub-nuimo-app/tasks/main.yml
+++ b/deployment/roles/hub-nuimo-app/tasks/main.yml
@@ -1,0 +1,84 @@
+---
+- name: create deployment location
+  become: true
+  file:
+    dest: "{{deploy_location}}"
+    owner: "{{deploy_user}}"
+    group: "{{run_user}}"
+    mode: "0755"
+    state: directory
+
+- name: create data location
+  become: true
+  file:
+    dest: "{{data_location}}"
+    owner: "{{run_user}}"
+    group: "{{run_user}}"
+    mode: "0750"
+    state: directory
+
+- name: give {{run_user}} access to bluez dbus
+  become: true
+  template:
+    dest: "/etc/dbus-1/system.d/bluetooth.conf"
+    owner: root
+    group: root
+    mode: "0644"
+    src: bluetooth.conf
+  notify: restart dbus
+
+- name: init virtualenv in deployment location
+  command: "{{python_interpreter}} -m venv venv"
+  args:
+    chdir: "{{deploy_location}}"
+    creates: "{{deploy_location}}/venv/bin/python"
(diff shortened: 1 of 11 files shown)


deployment/roles/hub-nuimo-app/tasks/main.yml
```diff
@@ -0,0 +1,84 @@
...
+  copy:
+    dest: "{{deploy_location}}/requirements.txt"
+    owner: "{{deploy_user}}"
+    group: "{{run_user}}"
+    mode: "0644"
+    content: |
+      pip==9.0.1
+      nuimo==0.1.0
+      websocket-client==0.40.0
+      https://github.com/grunskis/senic-hub-nuimo-app/archive/ftw.zip#egg=nuimo-app==0.0.1
```
tomster: we should make a release on our devpi server and install from there. this makes sure, that the package is correct and can 'survive' the packaging process

deployment/roles/hub-nuimo-app/tasks/main.yml
```diff
@@ -0,0 +1,84 @@
...
+
+- name: install application
+  pip:
+    virtualenv: "{{deploy_location}}/venv/"
+    requirements: "{{deploy_location}}/requirements.txt"
+    extra_args: "--pre --upgrade -i {{devpi_index_url}}"
+  notify: restart nuimo app
+
+- name: enable virtualenv to access the system packages
```
tomster: why not simply use ``--with-site-packages`` when creating the virtualenv?

deployment/roles/hub-nuimo-app/tasks/main.yml
grunskis: Ok. Once it lands in master I'll create a tag and a release. Actually I could use your help with releasing it...

deployment/roles/hub-nuimo-app/tasks/main.yml
grunskis: That was my initial approach. But then I noticed that this prevents `pip` to be installed within the `virtualenv`. So in the end I found that this is the easiest approach to achieve same thing with `venv` what was the usual behavior with `virtualenv`

deployment/roles/hub-nuimo-app/tasks/main.yml
tomster: yes, i would like to use the automatic versioning scheme, but we first need to review and merge https://github.com/getsenic/senic-hub-nuimo-app/pull/1 anyway, so we can do that there and then adjust the requirement here.

IOW not a showstopper

PR #22, opened 2017-02-28
Implement Wifi onboarding backend

tomster:


application/senic/nuimo_hub/views/setup_wifi.py
```diff
@@ -22,6 +30,11 @@ def scan_wifi_networks(request):
         return dict()
 
 
-@wifi_setup.post()
+@wifi_setup.post(renderer='json', schema=JoinWifiSchema)
```
larsblumberg: `renderer` attribute not necessary

application/senic/nuimo_hub/views/setup_wifi.py
```diff
@@ -22,6 +30,11 @@ def scan_wifi_networks(request):
...
 def join_network(request):
-    pass
+    run([
+        os.path.join(request.registry.settings['fs_bin'], 'join_wifi'),
```
larsblumberg: Needs to run as sudo

application/senic/nuimo_hub/tests/test_setup_wifi.py
```diff
@@ -0,0 +1,45 @@
+import pytest
+from mock import patch
```
grunskis: Shouldn't it be `from unittest.mock import patch`? https://docs.python.org/3.5/library/unittest.mock.html?highlight=mock#module-unittest.mock

application/senic/nuimo_hub/commands.py
```diff
@@ -15,33 +34,100 @@ def get_networks(devices=['wlan0', 'wlan1']):
...
+        run(['ifup', device], timeout=30)
+        # check, if we were successful:
+        status = run(['wpa_cli', 'status'], stdout=PIPE)
+        success = 'wpa_state=COMPLETED' in status.stdout.decode()
+    except TimeoutExpired:
+        success = False
+
+    if success:
+        # clean up after ourselves
+        if os.path.exists(ENTER_SETUP_FLAG):
```
grunskis: We could also consider using `pathlib.Path` since we're Python 3 only.

application/senic/nuimo_hub/commands.py
```diff
@@ -70,6 +70,7 @@ def activate_adhoc(device=DEFAULT_IFACE):
 @click.argument('device', default=DEFAULT_IFACE)
 def enter_wifi_setup(device=DEFAULT_IFACE):
     if not os.path.exists(ENTER_SETUP_FLAG):
+        os.click("Not entering wifi setup mode. %s not found" % ENTER_SETUP_FLAG)
```
larsblumberg: I would word it like that:
`os.click("Skipping Wi-Fi setup because %s does not exist" % ENTER_SETUP_FLAG)`

Can we rename `ENTER_SETUP_FLAG` to `WIFI_SETUP_FLAG` to make it more specific?

PR #23, opened 2017-03-01
Install or upgrade to BlueZ 5.44, closes #20

### LINKED ISSUE #20: Update playbooks to install/test Bluez 5.44
Readme in Linux SDK repo should be updated as well if 5.44 still works as expected

grunskis:
Install BlueZ 5.44 or upgrade if you have an older version already installed.


deployment/roles/ubuntu-bluez/tasks/from_src.yml
```diff
@@ -73,11 +73,19 @@
...
+
+- name: extract bluez version number from dict
+  set_fact:
+    custom_bluez_version: "{{ custom_bluez_version.stdout|default('') }}"
+
 - name: install sources
   command: make install
   args:
     chdir: "{{build_dir}}"
-    creates: /usr/local/bin/bluetoothctl
```
larsblumberg: Why is this line removed? Is it not necessary to check whether `/usr/local/bin/bluetoothctl` was created?

grunskis: It would skip the installation step if the file exists. In case of upgrade that's not what we want.

larsblumberg: Clever!

PR #32, opened 2017-03-03
Python 3.6 playbook

grunskis:
This would build Python 3.6 instead of 3.5

### DIFF
--- a/deployment/roles/python3-from-src/tasks/main.yml
+++ b/deployment/roles/python3-from-src/tasks/main.yml
@@ -1,8 +1,8 @@
 ---
-# A role to install Python 3.5 from sources
+# A role to install Python 3 from sources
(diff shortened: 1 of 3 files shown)


deployment/roles/python3-from-src/tasks/main.yml
```diff
@@ -76,11 +76,19 @@
     path: "{{build_dir}}/python"
   register: python_binary
 
+- name: get version of currently installed python3 version
```
tomster: i'm against making playbooks "too smart"... if we want python3.6 on  a specific host, we simply override the default version with `3.6.0` and apply this role as-is - done.

grunskis:
Ok. Let's keep using 3.4 for now.

PR #39, opened 2017-03-06
Add nuimo discovery and connect endpoint for on-boarding

larsblumberg:


application/senic/nuimo_hub/commands.py
```diff
@@ -135,3 +136,11 @@ def join_wifi(ssid, password, device=DEFAULT_IFACE):
         with open(ENTER_SETUP_FLAG, 'w') as flag:
             flag.write('FAILED')
         exit(1)
+
+
+@click.command(help='join a given wifi network (requires root privileges)')
```
tomster: looks like a copy&paste foobar

application/setup.py
```diff
@@ -80,5 +80,6 @@
         scan_wifi = senic.nuimo_hub.commands:scan_wifi
         enter_wifi_setup = senic.nuimo_hub.commands:enter_wifi_setup
         join_wifi = senic.nuimo_hub.commands:join_wifi
+        setup_nuimo = senic.nuimo_hub.commands:setup_nuimo
```
tomster: what's the reason to have this as CLT? because we want it callable via supervisord (IOW non-blocking)?

application/senic/nuimo_hub/commands.py
```diff
@@ -138,9 +138,15 @@ def join_wifi(ssid, password, device=DEFAULT_IFACE):
         exit(1)
 
 
-@click.command(help='join a given wifi network (requires root privileges)')
+@click.command(help="""Discover all nearby Nuimo controllers and connect to the first possible using
```
tomster: that's what i get for not looking ahead :)

application/senic/nuimo_hub/commands.py
```diff
@@ -35,7 +35,7 @@ def get_networks(devices=[DEFAULT_IFACE]):
 
 
 @click.command(help='scan the wifi interfaces for networks (requires root privileges)')
-@click.option('--config', '-c', default='development.ini', help='app configuration file')
+@click.option('--config', '-c', default='development.ini', type=click.Path(exists=True), help='app configuration file')
```
tomster: what does this do? make its presence (of the file) required?

application/senic/nuimo_hub/views/setup_nuimos.py
```diff
@@ -2,16 +2,40 @@
...
+
+connected_nuimos = Service(
+    name='connected_nuimos',
     path=service_path('setup/nuimos'),
     renderer='json',
     accept='application/json')
 
 
-@nuimo_setup.get()
+# TODO: Remove get()
```
tomster: why the `get` method here in the first place?

application/senic/nuimo_hub/views/setup_nuimos.py
```diff
@@ -23,13 +23,11 @@
...
 @nuimo_setup_starter.post(renderer='json')
 def start_nuimo_setup(request):
-    # TODO: Collect config file from some settings
     # TODO: Collect Bluetooth adapter name from some settings
-    # TODO: Should be started via supervisorctl to avoid multiple instances running
     run([
         'sudo',
         path.join(request.registry.settings['fs_bin'], 'setup_nuimo'),
-        '-c', '/srv/nuimo_hub/production.ini',
+        '-o', request.registry.settings['nuimo_mac_address_filepath'],
```
tomster: since both the writer and the reader of the file need to know where it is (and agree on its location) we've been using the pattern of passing around the location of the settings file instead of individual paths..

so the question here is simply: why?

application/senic/nuimo_hub/views/setup_nuimo.py
```diff
@@ -11,5 +12,10 @@
 
 
 @nuimo_setup.get()
-def discover_and_connect_nuimo(request):
-    return {'aa:bb:cc:dd:ee:ff': {'name': "Nuimo"}}
+def get_nuimo_mac_address(request):
```
tomster: this should definitely be covered by a test or two IMNSHO :)

application/senic/nuimo_hub/views/setup_nuimos.py
```diff
@@ -30,7 +30,7 @@ def bootstrap_nuimos(request):
         '-o', request.registry.settings['nuimo_mac_address_filepath'],
         'hci0'
     ])
-    return {}
+    return get_nuimo_mac_address(request)
```
tomster: this, too should definitely be covered by a test

application/senic/nuimo_hub/commands.py
larsblumberg: Yes, it will make the command line tool fail if the file isn't existent. In this case the file is needed. Not really part of the scope if this PR, but I came across it for the Nuimo discovery CLI tool, that priorly needed file presence, too.

application/setup.py
larsblumberg: The Nuimo discovery and connection was implemented a similar way as wifi scan. As we discussed this morning, we will run the Nuimo discovery code from the view endpoint instead. I suggest to do it with a different PR as the feature is implemented.

tomster: +1

application/senic/nuimo_hub/commands.py
tomster: neat click feature, that i didn't know about :)

PR #44, opened 2017-03-07
Nuimo discovery

larsblumberg:


application/development.ini
```diff
@@ -33,9 +33,9 @@ debug = true
 fs_bin = %(here)s/venv/bin/
 fs_wifi_networks = wifi_networks.json
 fs_enter_setup_flag = /var/local/WIFI_SETUP_REQUIRED
+nuimo_mac_address_filepath = nuimo_mac_address.txt
```
grunskis: We should rename all vars that have `fs_` prefix or stick to the convention until that's done

application/senic/nuimo_hub/nuimo_setup.py
```diff
@@ -0,0 +1,43 @@
...
+        self._manager = nuimo.ControllerManager(adapter_name)
+        self._manager.listener = self
+        self._controller = None
+
+    def discover_and_connect_controller(self):
+        # TODO: If there's a connected Nuimo, take it and don't run discovery
+        self._manager.start_discovery()
+        # Start D-Bus event loop. This call is blocking until the loop gets stopped.
+        # Will be stopped when a controller was connected (see below).
+        self._manager.run()
```
grunskis: What would happen if there are no Nuimos around or they are not in discovery mode? IMO you should allow Ctrl+C to kill the discovery or have a timeout

application/senic/nuimo_hub/views/setup_nuimos.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+connected_nuimos = Service(
+    name='connected_nuimos',
+    path=service_path('setup/nuimos'),
+    renderer='json',
+    accept='application/json')
+
+
+# TODO: Remove get()
```
grunskis: Reminder to remove GET :)

application/senic/nuimo_hub/views/setup_nuimos.py
```diff
@@ -0,0 +1,43 @@
...
+@nuimo_bootstrap.get()
+@nuimo_bootstrap.post(renderer='json')
+def bootstrap_nuimos(request):  # pragma: no cover,
+    # TODO: Collect Bluetooth adapter name from some settings
```
grunskis: There are already settings available through `{production,development}.ini` that're exposed to views through `request.registry.settings`.

application/senic/nuimo_hub/views/setup_nuimos.py
```diff
@@ -0,0 +1,43 @@
...
+connected_nuimos = Service(
+    name='connected_nuimos',
+    path=service_path('setup/nuimos'),
+    renderer='json',
+    accept='application/json')
+
+
+# TODO: Remove get()
+@nuimo_bootstrap.get()
+@nuimo_bootstrap.post(renderer='json')
```
grunskis: `renderer` is already set to `json` in the Service config. Unless you're overriding there's no need to specify again.

application/senic/nuimo_hub/views/setup_nuimos.py
```diff
@@ -0,0 +1,43 @@
...
+def bootstrap_nuimos(request):  # pragma: no cover,
+    # TODO: Collect Bluetooth adapter name from some settings
+    run([
+        'sudo',
+        path.join(request.registry.settings['fs_bin'], 'setup_nuimo'),
+        '-o', request.registry.settings['nuimo_mac_address_filepath'],
+        'hci0'
+    ])
+    return get_nuimo_mac_address(request)
```
grunskis: I find it awkward to call a view from another view. IMO would be better to create function that both views could call. It would also simplify unit test of that function b/c you won't need to create a `DummyRequest` to pass in...

application/senic/nuimo_hub/views/setup_nuimos.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+@connected_nuimos.get()
+def get_nuimo_mac_address(request):
+    nuimo_mac_address_filepath = request.registry.settings.get('nuimo_mac_address_filepath')
+    if not path.exists(nuimo_mac_address_filepath):
+        return []
```
grunskis: why not returning response in the same format as in the "happy path" like `{'connectedControllers': []}`?

application/senic/nuimo_hub/views/setup_nuimos.py
```diff
@@ -0,0 +1,43 @@
...
+    with open(nuimo_mac_address_filepath, 'r') as nuimo_mac_address_file:
+        mac_address = nuimo_mac_address_file.read().split('\n', 1)[0]
```
grunskis: IMO could be simplified: `f.readline().strip()`

application/senic/nuimo_hub/views/setup_nuimos.py
tomster: True, just like we render the wifi configurations in `/etc/network/interfaces.available` with the specific network interface names from ansbible, we can (and should) do the same for the bluetooth device

application/senic/nuimo_hub/views/setup_nuimos.py
tomster: actually, that's a not uncommon pattern. it's particularly useful (and appropriate IMHO) when you really just want to return the same response as the other view. `POST` or `PUT` views returning the same response as the dedicated `GET` view (as in this case) are the most common use case.

IMHO a view is just a 'helper' that receives a request as input and returns a response already. together with the added convenience of having access to the settings via the request i think the above solution is preferable.

grunskis: I don't see any advantage in passing around the request object instead of a simple string (`nuimo_mac_address_filepath`).

application/development.ini
larsblumberg: I'll provide a PR that removes `fs_` prefix if you're OK with it @grunskis @tomster

application/senic/nuimo_hub/views/setup_nuimos.py
larsblumberg: Fixed in @8521694040d3c00f02f0b2e06b0cc2b888de03af

application/senic/nuimo_hub/views/setup_nuimos.py
larsblumberg: Addressed in @ce1497f

application/senic/nuimo_hub/views/setup_nuimos.py
larsblumberg: Addressed in @72b5778

application/senic/nuimo_hub/views/setup_nuimos.py
larsblumberg: Addressed in @cd72a80

application/senic/nuimo_hub/views/setup_nuimos.py
larsblumberg: As both endpoints are supposed to return the same result I made one view calling the other

application/senic/nuimo_hub/nuimo_setup.py
larsblumberg: Addressed in @34cd26d

application/senic/nuimo_hub/views/setup_nuimos.py
larsblumberg: I'd like to keep it for now, makes manually testing the end point a little easier

larsblumberg:
@grunskis Addressed all your comments, can you please check again?

application/senic/nuimo_hub/__init__.py
```diff
@@ -1,4 +1,7 @@
 def main(global_config, **settings):        # pragma: no cover, tests have own app setup
+    import sys
+    from os import path
```
grunskis: Is there a reason not to put import in the beginning of the file?

application/senic/nuimo_hub/nuimo_setup.py
```diff
@@ -0,0 +1,54 @@
...
+
+    def __init__(self, adapter_name):
+        self._manager = nuimo.ControllerManager(adapter_name)
+        self._manager.listener = self
+        self._controller = None
+        self._timeout_timer = None
+
+    def discover_and_connect_controller(self, timeout=None):
+        # TODO: If there's a connected Nuimo, take it and don't run discovery
+        if timeout and timeout > 0:
```
grunskis: Could be just `if timeout:`

application/senic/nuimo_hub/__init__.py
larsblumberg: Addressed in @eb9dbbb

application/senic/nuimo_hub/nuimo_setup.py
larsblumberg: Addressed in @499f6c2

larsblumberg:
@grunskis Added a commit that might improve connection durations: Discovery is now stopped when we try to connect to a Nuimo.

grunskis:
👍 merge!!

PR #45, opened 2017-03-07
Run development version of hub on remote host using deployed stack

tomster:
@larsblumberg this is the work we did yesterday evening - i pushed it to the master of my own fork
@grunskis you might want to take a look, too, it might be useful for your workflow, too.


deployment/fab_pi3.py
```diff
@@ -2,6 +2,9 @@
 from os import path
 from fabric import api as fab
 from fabric.api import task, env
+from ploy.common import shjoin
+
+AV = None
```
grunskis: what `AV` stands for? :)

deployment/roles/hub-backend/templates/supervisor.conf
```diff
@@ -1,11 +1,16 @@
 [program:nuimo_hub]
```
grunskis: Maybe it would be better to have 2 different supervisor templates for dev/production instead of the `if/else`?

deployment/fab_pi3.py
tomster: ansible variables :) i kept it short, because in other scripts i use it all the time...

deployment/roles/hub-backend/templates/supervisor.conf
tomster: we discussed that and decided it's easier to maintain them and make sure they match if it's one template with a little bit of logic instead of two different templates in two different roles...

PR #47, opened 2017-03-08
Add endpoint to retrieve device details

grunskis:
Currently this endpoint only supports Philips Hue bridge and returns
all the lights know to the bridge.

### DIFF
--- a/application/senic/nuimo_hub/views/setup_devices.py
+++ b/application/senic/nuimo_hub/views/setup_devices.py
@@ -8,7 +8,7 @@ from pyramid.httpexceptions import HTTPBadRequest, HTTPNotFound
 from pyramid.response import FileResponse
 
 from ..config import path
-from ..device_discovery import PhilipsHueBridge, discover
+from ..device_discovery import PhilipsHueBridge, UnauthenticatedDeviceError, discover
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


application/senic/nuimo_hub/views/setup_devices.py
```diff
@@ -92,8 +87,38 @@ def devices_authenticate_view(request):
...
+
+    device_list_path = request.registry.settings['fs_device_list']
+    device = get_device(device_list_path, device_id)
+
+    data_location = request.registry.settings.get("fs_data_location", "/tmp")
+    bridge = PhilipsHueBridge(device["ip"], data_location)
+    if not bridge.is_authenticated():
+        raise HTTPBadRequest("Device not authenticated...")
+
+    return bridge.get_lights()
```
larsblumberg: Is `if not bridge.is_authenticated()` necessary? Doesn't also fail `bridge.get_lights()` if not authenticated?

application/senic/nuimo_hub/views/setup_devices.py
```diff
@@ -92,8 +87,38 @@ def devices_authenticate_view(request):
...
+    if not os.path.exists(device_list_path):
+        raise HTTPNotFound("Device discovery was not run...")
+
     with open(device_list_path, "r") as f:
         devices = json.loads(f.read())
 
-    return next((x for x in devices if x["id"] == device_id), None)
+    device = next((x for x in devices if x["id"] == device_id), None)
+    if device is None:
+        raise HTTPBadRequest("Device with id = {} not found...".format(device_id))
```
larsblumberg: `HTTPNotFound` 404 better in the case?

application/senic/nuimo_hub/views/setup_devices.py
grunskis: It does fail. I agree, it would be better to just call `get_lights` directly. I'll implement it...

application/senic/nuimo_hub/views/setup_devices.py
grunskis: Yep, you're right. Will update.

grunskis: Addressed in b85e502

application/senic/nuimo_hub/views/setup_devices.py
grunskis: Improved in 962032e

grunskis:
@larsblumberg addressed both points. Please take a look again.

larsblumberg:
Apart from the enum naming, LGTM

PR #51, opened 2017-03-09
Generate configuration for Home Assistant & Nuimo app

grunskis:
Command-line tool to generate HASS & Nuimo app config to be used as the last step of on-boarding.

No tests required since `command.py` is nicely excluded :P Not sure how much from this will stay in the end so probably no point to bother with tests...

### DIFF
--- a/application/senic/nuimo_hub/commands.py
+++ b/application/senic/nuimo_hub/commands.py
@@ -10,6 +10,12 @@ from .subprocess_run import run
 import wifi
 from pyramid.paster import get_app
 
+import configparser
+import re
+
+import yaml
+import requests
+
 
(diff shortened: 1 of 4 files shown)


application/senic/nuimo_hub/commands.py
```diff
@@ -135,3 +141,105 @@ def join_wifi(ssid, password, device=DEFAULT_IFACE):
         with open(ENTER_SETUP_FLAG, 'w') as flag:
             flag.write('FAILED')
         exit(1)
+
+
+@click.command(help='create configuration files for nuimo app & hass')
+@click.option('--config', '-c', default='development.ini', help='app configuration file')
```
larsblumberg: `type=click.Path(exists=True)`

PR #57, opened 2017-03-10
Show list of discovered devices

grunskis:
Runs device discovery and lists all the devices discovered


larsblumberg:
I would have named `_location` instead `_path` as with other existing settings to make clear it's a file path, but I'm okay with `location` for now.

application/senic/nuimo_hub/views/setup_config.py
```diff
@@ -0,0 +1,18 @@
...
+    path=path('setup/config'),
+    renderer='json',
+    accept='application/json',
+)
+
+
+@configuration_service.post()
+def configuration_create_view(request):
+    create_configuration_files_and_restart_apps_(request.registry.settings)
+    return {}
```
tomster: IMHO we should consistently either return `None` (the default) or an explicit status if a view has nothing meaningful to say. Any empty dict seems weird or rather too arbitrary.

frontend/src/SetupDevices.js
```diff
@@ -45,7 +45,7 @@ class SetupDevices extends Component {
       .then((response) => response.json())
       .then((devices) => {
         this.setState({ devices: devices })
-        this.discoverDevices()
+        //TODO: Run device discovery again as long as component is mounted
```
tomster: why?

grunskis: In case user plugs in new devices while discovering is already in process.

application/senic/nuimo_hub/views/setup_config.py
larsblumberg: Agree, `return None` or a `200` without body is good

grunskis: View doesn't have anything meaningful to say. I don't mind to return `None`. Updated in 06209c7

Issue #28, opened 2017-03-02
Add a WiFi status endpoint


larsblumberg:
1. **`/-/setup/wifi/connection`**
tells which WiFi network it has joined:

```
{
  ssid: "GetOfMyLawn",
  status: "connected" | "connecting" | "invalid-password" | "unavailable"
}
```
2. **`/-/setup/wifi/adhoc`**
tells if it has opened a adhoc network:
```
{
  "name": "Senic Hub",
  "status": "available" | "unavailable"
}
```

tomster:
implemented in #59

tomster:
this has been implemented

PR #59, opened 2017-03-13
#28 wifi status endpoint

tomster:


deployment/roles/hub-ubuntu/handlers/main.yml
```diff
@@ -21,9 +21,7 @@
     state: restarted
 
 - name: reload supervisord
-  service:
-    name: supervisor
-    state: restarted
+  raw: "supervisorctl reload"
```
grunskis: Note that there's also `supervisorctl` ansible module.

application/development.ini
```diff
@@ -31,14 +31,17 @@ pyramid.reload_templates = true
 pyramid.includes = pyramid_tm
 debug = true
 wifi_networks_path = /srv/nuimo_hub/data/wifi_networks.json
-wifi_setup_flag_path = /srv/nuimo_hub/data/WIFI_SETUP_REQUIRED
+wifi_setup_flag_path = /srv/nuimo_hub/data/wifi_setup_required
+wifi_adhoc_ssid = Setup Nuimo
```
larsblumberg: Let's call it: `Senic Setup`

application/senic/nuimo_hub/__init__.py
```diff
@@ -1,9 +1,10 @@
 import sys
 from os import path
-from .config import configure
 
 
 def main(global_config, **settings):        # pragma: no cover, tests have own app setup
     settings['bin_path'] = path.dirname(sys.executable)
+    from .config import configure
```
larsblumberg: Is this good style to import some modules within the method? Me as a less experienced Python dev would prefer all imports on the top

application/senic/nuimo_hub/commands.py
```diff
@@ -75,7 +73,9 @@ def activate_adhoc(device=DEFAULT_IFACE):
 
 @click.command(help='Activate the wifi-onboarding setup')
 @click.argument('device', default=DEFAULT_IFACE)
-def enter_wifi_setup(device=DEFAULT_IFACE):
+def enter_wifi_setup(config, device=DEFAULT_IFACE):
+    app = get_app(abspath(config))
+    ENTER_SETUP_FLAG = app.registry.settings['wifi_setup_flag_path']
```
larsblumberg: Consistency: `ENTER_SETUP_FLAG` –> `WIFI_SETUP_FLAG`, because of `wifi_setup_flag_path `

application/senic/nuimo_hub/tests/conftest.py
```diff
@@ -32,12 +32,15 @@ def settings():
         'debug': False,
         'bin_path': asset_path('bin'),
         'wifi_networks_path': asset_path('wifi_networks.json'),
-        'wifi_setup_flag_path': asset_path('NUIMO_SETUP_REQUIRED'),
+        'wifi_setup_flag_path': asset_path('nuimo_setup_required'),
+        'wifi_adhoc_ssid': 'Setup Nuimo',
```
larsblumberg: `Senic Setup`

application/senic/nuimo_hub/tests/test_setup_wifi.py
```diff
@@ -30,16 +30,67 @@ def mocked_run(request):
...
+        status='unavailable'
+    )
+
+
+@pytest.fixture
+def adhoc_url(route_url):
+    return route_url('wifi_adhoc')
+
+
+def test_get_adhoc_wifi(settings, browser, adhoc_url):
```
larsblumberg: `test_get_adhoc_wifi_available`?

application/senic/nuimo_hub/tests/test_setup_wifi.py
```diff
@@ -30,16 +30,67 @@ def mocked_run(request):
...
+    )
+
+
+@pytest.fixture(scope='function')
+def no_wifi_setup_mode(settings):
+    settings['wifi_setup_flag_path'] = '/no/such/file'
+    return settings
+
+
+def test_get_no_adhoc_wifi(no_wifi_setup_mode, settings, browser, adhoc_url):
```
larsblumberg: `test_get_adhoc_wifi_unavailable`?

deployment/roles/hub-ubuntu/handlers/main.yml
tomster: yes, but that does not support reloading `supervisord` unfortunately, and using the `service` module (which delegates to `systemctl`) randomly fails with a `systemd`-specific error. hence the workaround

grunskis: If I understood the docs correctly then the update/add of the config file happens automatically depending on the `state` variable.

I've seen those "random" failures too and AFAICT they are not that random. Sometimes the supervisor config was wrong due to the mixed paths. Some were fixed by splitting up the location variables but there might be other problems hidden...

application/development.ini
tomster: Done in a945616

application/senic/nuimo_hub/__init__.py
tomster: Normally, yes. but in global `__init__.py` such as here it is good practice to avoid imports on the top, because they would be triggered by anybody importing your package.

One specific reason to be conservative here is (unfortunately) that premature imports of your own modules can mess with coverage reporting.

so, yes, it's not nice or good style but here it's better to err on the safe side.

application/senic/nuimo_hub/commands.py
tomster: addressed in d87ef52

application/senic/nuimo_hub/__init__.py
larsblumberg: I see – thanks for the explanation!

application/senic/nuimo_hub/commands.py
```diff
@@ -75,9 +75,9 @@ def activate_adhoc(device=DEFAULT_IFACE):
 @click.argument('device', default=DEFAULT_IFACE)
 def enter_wifi_setup(config, device=DEFAULT_IFACE):
     app = get_app(abspath(config))
-    ENTER_SETUP_FLAG = app.registry.settings['wifi_setup_flag_path']
-    if not os.path.exists(ENTER_SETUP_FLAG):
-        click.echo("Not entering wifi setup mode. %s not found" % ENTER_SETUP_FLAG)
+    WIFI_SETUP_FLAG = app.registry.settings['wifi_setup_flag_path']
```
larsblumberg: Ah, we had a misunderstanding here. The path setting is called `wifi_setup_flag_path`, hence the constant should be named the same:

    WIFI_SETUP_FLAG_PATH = app.registry.settings['wifi_setup_flag_path']

PR #61, opened 2017-03-14
Improved Philips Hue bridge discovery

grunskis:
Before there was an intermediate state file created with Philips Hue authentication info. Then this file was used during the completion step to generate configuration for Home Assistant.

Now the intermediate state file is not created anymore and the Home Assistant configuration is created directly.

### DIFF
--- a/application/senic/nuimo_hub/device_discovery.py
+++ b/application/senic/nuimo_hub/device_discovery.py
@@ -1,6 +1,6 @@
+import functools
 import json
 import logging
-import os.path
 import re
 
 from enum import IntEnum
@@ -20,8 +20,15 @@ class UnauthenticatedDeviceError(Exception):
     message = "Device not authenticated..."
 
 
+class UpstreamError(Exception):
+    def __init__(self, error_type=None):
+        self.error_type = error_type
+        self.message = "Received error from the upstream device! Type: {}".format(self.error_type)
+
+
 class PhilipsHueBridgeError(IntEnum):
     unauthorized = 1
+    button_not_pressed = 101
 
 
 logger = logging.getLogger(__name__)
@@ -80,13 +87,26 @@ def extract_philips_hue_bridge_ip(device_info):
         return result.group(1)
 
 
+class username_required:
+    def __init__(self, method):
+        self.method = method
+
+    def __call__(self, instance, *args, **kwargs):
+        if instance.username is None:
+            raise UnauthenticatedDeviceError()
+
+        return self.method(instance, *args, **kwargs)
+
+    def __get__(self, instance, instancetype):
+        return functools.partial(self.__call__, instance)
+
+
 class PhilipsHueBridge:
-    def __init__(self, ip_address, data_location):
+    def __init__(self, ip_address, username=None):
         self.bridge_url = "http://{}/api".format(ip_address)
-        self.state_file_path = os.path.join(data_location, ip_address)
+        self.username = username
         self.app_name = "senic hub#192.168.1.12"  # TODO get real IP of the hub
 
-        self._username = None
         self._http_session = requests.Session()
 
     def _request(self, url, method="GET", payload=None, timeout=5):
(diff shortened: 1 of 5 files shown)


application/senic/nuimo_hub/device_discovery.py
```diff
@@ -123,16 +147,12 @@ def authenticate(self):
...
-            self._username = response["success"]["username"]
-
-            with open(self.state_file_path, "w") as f:
-                json.dump({"devicetype": self.app_name, "username": self._username}, f)
-
+            if response:
+                self.username = response["success"]["username"]
         except UnauthenticatedDeviceError:
-            pass  # consume the exception
+            self.username = None
```
larsblumberg: What's the benefit of resetting the username to `None`? What if we leave it as it was, knowing that it's (potentially) not valid any longer

application/senic/nuimo_hub/device_discovery.py
```diff
@@ -80,13 +87,26 @@ def extract_philips_hue_bridge_ip(device_info):
         return result.group(1)
 
 
+class authentication_required:
```
larsblumberg: This decorator only tests for username != None. What if we rename it to `username_required` to explicitly express that requirement?

application/senic/nuimo_hub/views/setup_devices.py
```diff
@@ -79,15 +84,35 @@ def devices_authenticate_view(request):
     if device["type"] != "philips_hue":
         raise HTTPBadRequest("Device doesn't require authentication...")
 
-    data_location = request.registry.settings["data_path"]
-    bridge = PhilipsHueBridge(device["ip"], data_location)
+    config = read_json(request.registry.settings["hass_phue_config_path"], {})
+    if device["ip"] in config:
+        username = config[device["ip"]]["username"]
+    else:
+        username = None
```
larsblumberg: If you find it readable, we could also type:
```python
    username = (config.get(device["ip"]) or {}).get("username")
```

application/senic/nuimo_hub/views/setup_devices.py
```diff
@@ -79,15 +84,35 @@ def devices_authenticate_view(request):
...
     if not bridge.is_authenticated():
-        logger.debug("Trying to authenticate Philips Hue bridge %s", device["ip"])
-        bridge.authenticate()
+        username = bridge.authenticate()
+
+    if username:
+        config[device["ip"]] = {"username": username}
+    else:
+        if device["ip"] in config:
+            del config[device["ip"]]
```
larsblumberg: Not necessarily more readable but avoids `if`:

    config.pop(device["ip"], None)

application/senic/nuimo_hub/views/setup_devices.py
```diff
@@ -105,13 +130,21 @@ def devices_details_view(request):
     device_list_path = request.registry.settings['devices_path']
     device = get_device(device_list_path, device_id)
 
-    data_location = request.registry.settings["data_path"]
-    bridge = PhilipsHueBridge(device["ip"], data_location)
+    config = read_json(request.registry.settings["hass_phue_config_path"])
+    if config and device["ip"] in config:
+        username = config[device["ip"]]["username"]
+    else:
+        username = None
```
larsblumberg: Former suggestions might be valid, but now I might be exaggerating. I like it though because it reduces reading efforts :)
```python
username = ((config or {}).get(device["ip"]) or {}).get("username")
```

application/senic/nuimo_hub/device_discovery.py
grunskis: If I set it to `None` then subsequent requests don't have to try reaching bridge to fail and fail fast.

larsblumberg: Nice! 👍

application/senic/nuimo_hub/views/setup_devices.py
grunskis: Good point. I don't like `del key` syntax either. Will update...

application/senic/nuimo_hub/device_discovery.py
grunskis: Updated 328f5f3

application/senic/nuimo_hub/views/setup_devices.py
grunskis: This and all previous similar changes in 687b302

PR #66, opened 2017-03-14
Show name of Philips Hue bridges & Sonos speakers

### LINKED ISSUE #46: Use MAC address as a device ID for device discovery
Currently a sequential number is used as an ID which is not very reliable in the long run. If possible a MAC address or some other identifier should be used that doesn't change over time.

grunskis:
Also adds polling of authentication status, so that there's more time
for pressing the button Philips Hue bridge

Also closes #46

### DIFF
--- a/application/senic/nuimo_hub/tests/test_device_discovery.py
+++ b/application/senic/nuimo_hub/tests/test_device_discovery.py
@@ -2,6 +2,8 @@ from unittest.mock import MagicMock
 
 from pytest import fixture
 
+import responses
+
 from senic.nuimo_hub.device_discovery import discover, extract_philips_hue_bridge_ip, make_device_description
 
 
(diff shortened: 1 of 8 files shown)


grunskis:
Fixing tests :man_facepalming:

application/senic/nuimo_hub/tests/test_device_discovery.py
```diff
@@ -15,35 +17,48 @@ def sonos_device_info():
...
 def test_make_device_description_sonos(sonos_device_info):
+    response_xml = """
+<root xmlns="urn:schemas-upnp-org:device-1-0">
+  <device>
+    <friendlyName>192.168.1.42 Foo Bar</friendlyName>
+    <UDN>uuid:RINCON_123</UDN>
+    <roomName>Foo Bar</roomName>
+  </device>
+</root>
+"""
```
larsblumberg: The xml should still be valid if it's properly indented, no?

larsblumberg: I.e. adding spaces to each line

grunskis: I could add spaces. Can you propose a way to indent this?

frontend/src/SetupDevices.js
```diff
@@ -38,21 +41,43 @@ class SetupDevices extends Component {
...
       })
       .catch((error) => console.error(error))
   }
+
+  authenticateDevice(device) {
+    fetch('/-/setup/devices/' + device.id + '/authenticate', {method: 'POST'})
+      .then((response) => response.json())
+      .then((response) => {
+        if (!response.authenticated) {
+          var timer = setTimeout(this.authenticateDevice.bind(this, device), this.authPollInterval)
```
larsblumberg: Let's use `let` for declaring variables instead of `var`, it has a couple of advantages such as visibility only in scope, no re-declaration, and only accessible after its declaration

application/senic/nuimo_hub/tests/test_device_discovery.py
larsblumberg:
```python
def test_make_device_description_sonos(sonos_device_info):
 +    response_xml = """
 +        <root xmlns="urn:schemas-upnp-org:device-1-0">
 +          <device>
 +            <friendlyName>192.168.1.42 Foo Bar</friendlyName>
 +            <UDN>uuid:RINCON_123</UDN>
 +            <roomName>Foo Bar</roomName>
 +          </device>
 +        </root>
 +        """
```

larsblumberg: Like this? Not sure where to put the `"""` though

grunskis: Done 5faa2e6

grunskis:
@larsblumberg Did the reindentation for XML. Let me know if this is fine, I would like to squash this before merging...

PR #68, opened 2017-03-15
Continuous integration or Toxify all the things!!!

tomster:
This change introduces building and uploading the package for each commit on or merge into the master branch onto our devpi instance at https://pypi.senic.com/

Each developer fork has its own index, so there shouldn't be any versioning clashes.


tomster:
Note, that merging this PR should automatically result in a package appearing at https://pypi.senic.com/senic/master (which is currently still empty)

tomster:
Also, maybe one of you has an idea how to best implement slack notifications?

larsblumberg:
I wouldn't be surprised if there was a Slack integration for Travis, we can pair on that

grunskis:
LGTM! I might have more comments after trying it out.

PR #69, opened 2017-03-16
Device discovery daemon

grunskis:
Converts device discovery into a daemon that's managed by supervisord. Discovery is running every minute and is updating the "DB" with newly discovered devices. Frontend is polling the GET endpoint for new devices. It's also possible to restart the discovery by POSTing to `/-/setup/devices/discover` which will trigger a restart of supervisor program through their RPC interface.

`discover_devices` command doesn't have any tests, but probably should. Let me know what you think...

Includes also other fixes and improvements that I've stumbled upon while working on this. Sorry about that.


application/senic/nuimo_hub/commands.py
```diff
@@ -228,3 +242,55 @@ def get_component_for_device(device):
...
+        devices = []
+
+    while True:
+        now = datetime.utcnow()
+
+        # remove discovery time from known devices, to simplify logic
+        # of finding only the new devices
+        known_devices = deepcopy(devices)
+        for device in known_devices:
+            device.pop("timeOfDiscovery", None)
```
larsblumberg: `discoveryDate` or `discoveryTimestamp` as an easier name?

deployment/roles/hub-ubuntu/tasks/main.yml
```diff
@@ -137,6 +147,7 @@
   with_items:
     - scan_wifi
     - wifi_setup
+    - discover_devices
```
larsblumberg: If would opt for using nouns for command line tools, i.e. `device_discovery`

frontend/src/SetupDevices.js
```diff
@@ -5,6 +5,9 @@ import { Link } from 'react-router'
 import './SetupDevices.css'
 
 class SetupDevices extends Component {
+  devicePollInterval = 5000  // 5 seconds
+  devicePollTimer = null
```
larsblumberg: Since we're polling for multiple devices, can we name it `devicesPollInterval` and `devicesPollTimer`?

application/senic/nuimo_hub/commands.py
```diff
@@ -228,3 +242,55 @@ def get_component_for_device(device):
...
+
+        discovered_devices = discover(known_devices=known_devices)
+
+        if discovered_devices:
+            for device in discovered_devices:
```
larsblumberg: To avoid `if` and thus one nesting level, we could type:

    for device in (discovered_devices or [])

application/senic/nuimo_hub/commands.py
```diff
@@ -228,3 +242,55 @@ def get_component_for_device(device):
...
+                # make sure we get updates for devices we already had discovered before
+                devices = [x for x in devices if x["id"] != device["id"]]
```
larsblumberg: I'm fine with one letter vars in comprehensions but can we use the initial letter?, i.e. `for d in devices`

application/senic/nuimo_hub/commands.py
```diff
@@ -228,3 +242,55 @@ def get_component_for_device(device):
...
+                device["timeOfDiscovery"] = str(now)
+                devices.append(device)
+
+            with open(devices_path, 'w') as f:
+                json.dump(devices, f)
+
+        next_scan = now + timedelta(seconds=scan_interval_seconds)
+        logging.info("Next device discovery run scheduled for {}".format(next_scan))
+        while datetime.utcnow() < next_scan:
+            time.sleep(1)
```
larsblumberg: Alternatively, can we sleep the remaining time without using `while`? I.e.

    time.sleep(next_scan - datetime.utcnow())

application/senic/nuimo_hub/device_discovery.py
```diff
@@ -35,7 +36,16 @@ class PhilipsHueBridgeError(IntEnum):
 logger = logging.getLogger(__name__)
 
 
-def discover(discovery_class=NetworkDiscovery):
+def discover(discovery_class=NetworkDiscovery, known_devices=None):
```
larsblumberg: It's better do default `known_devices` to `[]` and not allow `None`, then we don't need to test for `known_devices is None` as you do later

application/senic/nuimo_hub/device_discovery.py
```diff
@@ -50,6 +60,10 @@ def discover(discovery_class=NetworkDiscovery):
         for device_info in netdisco.get_info(device_type):
             device_description = make_device_description(device_type, device_info)
             if device_description:
+                if known_devices and device_description in known_devices:
+                    continue
+
+                logger.info("Discovered new %s device with ip %s", device_type, device_description["ip"])
                 devices.append(device_description)
```
larsblumberg:
```python
    device_descriptions = [make_device_description(device_type, d) for d in netdisco.get_info(device_type)]
    new_device_descriptions = [d for d in device_descriptions if d not in known_devices]
    logger.info(...)
    devices.extend(new_device_descriptions)
```

deployment/roles/hub-ubuntu/templates/discover_devices_supervisor.conf
```diff
@@ -0,0 +1,10 @@
+[program:discover_devices]
+command = {{backend_deploy_location}}/venv/bin/discover_devices -c {{backend_deploy_location}}/production.ini
+autostart=true
+autorestart=true
+directory={{ backend_deploy_location }}
+stdout_logfile={{ backend_data_location }}/discover_devices.log
+redirect_stderr=true
+user={{ run_user }}
+startsecs=3
+environment=LC_ALL=C.UTF-8, LANG=C.UTF-8
```
larsblumberg: If you rename the cmd line tool to `device_discovery`, this .conf file also needs to be renamed

nuimo_app/senic/nuimo_app/__main__.py
```diff
@@ -24,24 +24,24 @@ def main(config_file_path=DEFAULT_CONFIG_FILE_PATH):
...
+    ha_url = config.get("ha_api_url", "localhost:8123")
     ha_api = HAListener("ws://{}".format(ha_url))
     ha_api.start()
 
     nuimo_app = None
     controller_mac_address = config.get("controller_mac_address")
     if controller_mac_address:
-        ble_adapter_name = config.get("ble_adapter_name", DEFAULT_BLE_ADAPTER_NAME)
-        nuimo_app = NuimoApp(controller_mac_address, ha_api, ble_adapter_name)
+        nuimo_app = NuimoApp(ha_api, controller_mac_address, manager)
```
larsblumberg: Shouldn't we let `NuimoApp` create the `ControllerManager` internally and continue to pass `ble_adapter_name` to `NuimoApp()`?

application/senic/nuimo_hub/commands.py
grunskis: Wouldn't that block the main thread?

larsblumberg: Doesn't it block it anyway already?

nuimo_app/senic/nuimo_app/__main__.py
grunskis: Maybe that could work. I remember that there was a problem that Ctrl+C wasn't responding otherwise, but that might have changed. I'll look into this...

application/senic/nuimo_hub/commands.py
larsblumberg: Alternatively:

    discovered_devices = discover(known_devices=known_devices) or []

larsblumberg: And alternatively, even better :) `discover` method never returns `None`, always an array (or dict, dunno)

application/senic/nuimo_hub/commands.py
grunskis: `x` is the standard name. like `i` in C for looping. IMHO it also looks better in most cases than any other letter :D

application/senic/nuimo_hub/commands.py
grunskis: Nope, it doesn't. If it would then supervisor wouldn't be able to restart the program... I'll check how it behaves with long sleeps...

grunskis: Yep, killing still works also when sleeping the whole period, fa3313e

frontend/src/SetupDevices.js
grunskis: Renamed b26aa64

application/senic/nuimo_hub/commands.py
grunskis: Renamed afd5342

application/senic/nuimo_hub/commands.py
grunskis: It actueally doesn't return `None`

application/senic/nuimo_hub/commands.py
grunskis: Refactored and took care of this. By popular demand will avoid using `x` as var name ;)

application/senic/nuimo_hub/commands.py
grunskis: In the end renamed the field to `discovered` to match naming scheme of other fields which don't include type as part of name

nuimo_app/senic/nuimo_app/__main__.py
grunskis: Did as you suggested here 0c59d6d Thanks!

grunskis:
@larsblumberg addressed all the things besides rename of the daemon. if we decide to rename it then we should use the same naming pattern for all daemons/programs that we have.

Let me know if you have any other comments, otherwise I'll squash...

larsblumberg:
LGTM! Nice PR!

PR #72, opened 2017-03-17
Improve nuimo discovery

larsblumberg:
- Fixes one issue where an instance variable wasn't defined. 
- Now nuimo lib >= 0.3.0 is required, it fixes mainly that D-Bus signal connections weren't removed

### DIFF
--- a/application/setup.py
+++ b/application/setup.py
@@ -38,7 +38,7 @@ setup(
         'click',
         'colander',
         'cornice<2.0',
-        'nuimo>=0.2.3',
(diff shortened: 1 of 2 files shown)


application/setup.py
```diff
@@ -38,7 +38,7 @@
         'click',
         'colander',
         'cornice<2.0',
-        'nuimo>=0.2.3',
+        'nuimo>=0.3.0',
```
grunskis: Not sure how useful is `>=` pinning.

grunskis:
On comment about pinning, otherwise 👍

PR #77, opened 2017-03-20
Always return up-to-date authentication status

grunskis:
Device list endpoint was returning incorrect authentication status sometimes. Fronted was not being updated correctly either.


larsblumberg:
LGTM

PR #78, opened 2017-03-22
Simplify nuimo app

grunskis:
There was support for controlling multiple Home Assistant devices (aka entities) in Nuimo App. I removed this support to simplify code + it's not really needed at this point.

### DIFF
--- a/senic_hub/nuimo_app/components.py
+++ b/senic_hub/nuimo_app/components.py
@@ -97,19 +97,19 @@ class PhilipsHue(Component):
         """
         delta = int(value / 1800 * self.MAX_BRIGHTNESS_VALUE)
 
-        new_value = min([max([0, self.brightness + delta]), self.MAX_BRIGHTNESS_VALUE])
+        new_value = min(max(self.brightness + delta, 0), self.MAX_BRIGHTNESS_VALUE)
         logger.debug("%s brightness current: %s delta: %s new: %s", self.entity_id,
                      self.brightness, delta, new_value)
 
         self.brightness = new_value
 
-        # turn off bulbs if brightness has reached 0
         args = dict(transition=0)
-        if new_value > 1:
+        if new_value:
             service = "turn_on"
             args["brightness"] = new_value
             icon = icons.light_bar(self.MAX_BRIGHTNESS_VALUE, new_value)
         else:
+            # turn off bulbs if brightness has reached 0
             service = "turn_off"
             icon = icons.POWER_OFF
 
@@ -118,8 +118,7 @@ class PhilipsHue(Component):
 
     def swipe_left(self):
         led_cfg = LEDMatrixConfig(icons.LETTER_W)
-        brightness = max(self.brightness.values())
-        return self.make_action("turn_on", led_cfg, rgb_color=COLOR_WHITE_RGB, brightness=brightness)
+        return self.make_action("turn_on", led_cfg, rgb_color=COLOR_WHITE_RGB, brightness=self.brightness)
 
     def swipe_right(self):
         led_cfg = LEDMatrixConfig(icons.SHUFFLE)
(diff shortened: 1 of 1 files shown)


nuimo_app/senic/nuimo_app/components.py
```diff
@@ -51,43 +51,33 @@ class PhilipsHue(Component):
...
-        self.is_light_on = all(x == "on" for x in self.state.values())
+        logger.debug("%s state: %s brightness: %s", self.entity_id, self.state, self.brightness)
 
     def button_press(self):
         """
         Toggle the state of all lights.
 
         """
-        if self.is_light_on:
+        if self.state == "on":
```
larsblumberg: No need to change now, but I suggest that we [use quotation marks as described here](http://stackoverflow.com/a/56190/543875). TL;DR: Use single quotes for symbols. Use double quotes for strings that are supposed to be read by the user on the screen or by the programmer on the console.

nuimo_app/senic/nuimo_app/components.py
```diff
@@ -107,21 +97,18 @@ def rotation(self, value):
         """
         delta = int(value / 1800 * self.MAX_BRIGHTNESS_VALUE)
 
-        for entity_id, brightness in self.brightness.items():
-            new_value = min([max([0, brightness + delta]), self.MAX_BRIGHTNESS_VALUE])
-            logger.debug("%s brightness current: %s delta: %s new: %s", entity_id,
-                         brightness, delta, new_value)
+        new_value = min([max([0, self.brightness + delta]), self.MAX_BRIGHTNESS_VALUE])
```
larsblumberg: Better readable if min and max values stand next to each other:

    new_value = max(0, min(self.MAX_BRIGHTNESS_VALUE, self.brightness + delta))

or

    new_value = min(max(self.brightness + delta, 0), self.MAX_BRIGHTNESS_VALUE)

Is it necessary to mix `[]` and `()`?

nuimo_app/senic/nuimo_app/components.py
```diff
@@ -107,21 +97,18 @@ def rotation(self, value):
...
 
-            self.brightness[entity_id] = new_value
-
-        max_value = max(self.brightness.values())
+        self.brightness = new_value
 
         # turn off bulbs if brightness has reached 0
         args = dict(transition=0)
-        if max_value > 1:
+        if new_value > 1:
```
larsblumberg: If it's intentionally not `> 0` then I'd like to read a comment here why `> 1` is necessary

nuimo_app/senic/nuimo_app/nuimo_app.py
```diff
@@ -152,26 +152,20 @@ def call_service_callback(entity_id, response):
             logger.debug("service_call response for %s:", entity_id)
             logger.debug(pformat(response))
 
-            status = response["success"]
-            action.entity_updated(entity_id, status)
+            if response["success"]:
+                matrix_config = action.led_matrix_config
+            else:
+                matrix_config = LEDMatrixConfig(icons.ERROR)
```
larsblumberg: Don't know if you like it:

    matrix_config = action.led_matrix_config if response["success"] else LEDMatrixConfig(icons.ERROR)

nuimo_app/senic/nuimo_app/components.py
grunskis: It's not really mixing, but making a list and choosing min/max from that. Just checked that in modern python versions don't require to have a list as a parameter to min/max. Will update...

nuimo_app/senic/nuimo_app/nuimo_app.py
grunskis: For me it's hard to read long lines and not obvious that it's a condition.

grunskis:
@larsblumberg addressed your feedback

PR #79, opened 2017-03-22
Reconnect to Home Assistant if connection is lost

grunskis:
Based on branch used in https://github.com/getsenic/nuimo-hub-backend/pull/78

### DIFF
--- a/senic_hub/nuimo_app/__init__.py
+++ b/senic_hub/nuimo_app/__init__.py
@@ -5,7 +5,9 @@ from pprint import pformat
 
 from nuimo import (Controller, ControllerListener, ControllerManager, Gesture)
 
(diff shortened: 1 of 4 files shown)


nuimo_app/senic/nuimo_app/components.py
```diff
@@ -27,7 +27,10 @@ def state_changed(self, state):
         logger.debug("state_changed:")
         logger.debug(pformat(state))
 
-        new_state = state["data"]["new_state"]
+        if "data" in state:
+            new_state = state["data"]["new_state"]
+        else:
+            new_state = state
```
larsblumberg: In case you like that, too :)

    new_state = state["data"]["new_state"] if "data" in state else state

PR #81, opened 2017-03-22
Merging backend, frontend and nuimo_app into a single package

tomster:


tomster:
note, that this contains #78

PR #83, opened 2017-03-23
Make sure users exist before using them in a role

grunskis:
Running any of the roles for the first time would end with an error unless you had run `hub-backend` role first. 

I created a new role that creates users/groups & made it a dependency for any role that's using any of them (actually all of them)


larsblumberg:
LGTM

PR #95, opened 2017-03-29
Refactor device discovery

grunskis:
* Device descriptions are fetched separately and the rest of the details are
  added later.
* Philips Hue bridge authentication info is written to a separate file
  for each discovered bridge (this is actually how Home Assistant expects it to be) 
* Only the authenticated devices are added to the configuration of
  nuimo app and home assistant


senic_hub/backend/commands.py
```diff
@@ -270,10 +282,45 @@ def discover_devices(config):
...
+
+
+def add_authentication_status(devices, settings):
+    for device in devices:
+        if not device["authenticationRequired"]:
+            device["authenticated"] = True
+            continue
+
+        if device["type"] != "philips_hue":
+            continue
```
larsblumberg: Is this test still necessary? `device["authenticationRequired"]` is already tested.

senic_hub/backend/tests/test_device_discovery.py
```diff
@@ -1,65 +1,68 @@
...
 
-from pytest import fixture
+from pytest import fixture, raises
 
 import responses
 
-from senic_hub.backend.device_discovery import (DISCOVERY_TIMESTAMP_FIELD, discover, discover_and_update_devices,
-                                              extract_philips_hue_bridge_ip, make_device_description)
+from senic_hub.backend.device_discovery import (DISCOVERY_TIMESTAMP_FIELD, UpstreamError,
+                                                discover, discover_devices, get_device_description)
```
larsblumberg: I'm okay with that formatting but I can tell you that @tomster and me would format it like that:

    from senic_hub.backend.device_discovery import (
        DISCOVERY_TIMESTAMP_FIELD, UpstreamError, discover, discover_devices, get_device_description)

senic_hub/backend/commands.py
grunskis: Yeah, the code that follows is philips hue specific.

senic_hub/backend/tests/test_device_discovery.py
grunskis: Updated in 3bf8e82b47d3ae27fd704fa8f3de8a4d838d1f3c

PR #98, opened 2017-03-30
Run discovery more often

grunskis:
Run device discovery every 15 seconds during onboarding and then stop it after we're done.

I'm excluding `network_discovery.py ` from test coverage check. It's mostly copy/paste from netdisco. IMO we should rewrite it to make it simpler and then test it, but for now it works as it is.

### DIFF
--- a/senic_hub/backend/device_discovery.py
+++ b/senic_hub/backend/device_discovery.py
@@ -6,9 +6,9 @@ import xml.etree.ElementTree as ET
 from copy import deepcopy
 from enum import IntEnum
 
-from netdisco.discovery import NetworkDiscovery
-
 import requests
+
+from .network_discovery import NetworkDiscovery
 
 
 SUPPORTED_DEVICES = [
@@ -77,13 +77,10 @@ def discover(discovery_class=NetworkDiscovery):
 
     devices = []
 
-    netdisco = discovery_class()
+    netdisco = discovery_class(limit_discovery=SUPPORTED_DEVICES)
     netdisco.scan()
 
     for device_type in netdisco.discover():
-        if device_type not in SUPPORTED_DEVICES:
-            continue
-
         for device_info in netdisco.get_info(device_type):
             device_description = get_device_description(device_type, device_info)
             logger.info("Discovered %s device with ip %s", device_type, device_description["ip"])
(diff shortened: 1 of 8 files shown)


senic_hub/backend/device_discovery.py
```diff
@@ -77,13 +77,10 @@ def discover(discovery_class=NetworkDiscovery):
 
     devices = []
 
-    netdisco = discovery_class()
+    netdisco = discovery_class(limit_discovery=SUPPORTED_DEVICES)
```
larsblumberg: It's not clear to me that `limit_discovery` takes a list of supported devices. It sounds like boolean variable to me. I suggest `device_filter`.

senic_hub/backend/network_discovery.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+class NetworkDiscovery(object):
+    """
+    Based on netdisco.discovery.NetworkDiscovery. This class only does
+    discovery using SSDP & nUPNP which covers all the devices we want
+    to support at the moment.
+
+    """
+    def __init__(self, limit_discovery=[]):
```
larsblumberg: What about creating a PR to netdisco that introduces this optional parameter?

larsblumberg: Default value should be `None`, not an empty array. Empty array should not include any device.

senic_hub/backend/device_discovery.py
grunskis: It's not really a filter. It's a list of devices to search for. Would `discover_devices` or `device_whitelist` work better as a name for it?

larsblumberg: `whitelist` sounds good!

senic_hub/backend/network_discovery.py
larsblumberg: (updated comment, was hard to understand)

grunskis: Yep, that what the intention. I'll remove the default value, which would make it more clear.

senic_hub/backend/device_discovery.py
grunskis: Addressed in f631c15

PR #101, opened 2017-04-04
Use two wifi adapters for on-boarding: one for ad-hoc, the other for infrastructure

larsblumberg:
* moved most wifi setup commands into `wifi_setup` cmd line tool
* provisioning machine with 2 wlan adapters, configurable via ploy.conf
* read wifi connection state via `wpa_cli` to get rid of connection state json file
* drop support for single wlan adapter based on-boarding
* don't control `dhcpd` via supervisorctl but let system take care of it, cause it's quiet integrated already (at least on a raspbian)


deployment/fab_pi3.py
```diff
@@ -6,16 +6,17 @@
...
-interface {eth_iface}
-static ip_address={eth_ip}/{eth_netmask}
-static routers={eth_gateway}
-static domain_name_servers={eth_dns}
+eth_interface = """
+auto {eth_iface}
+iface {eth_iface} inet static
+  address {eth_ip}
+  netmask {eth_netmask}
+  gateway {eth_gateway}
```
grunskis: We found out yesterday that this would break the name resolution if we don't have ethernet plugged in

deployment/fab_pi3.py
```diff
@@ -32,18 +33,24 @@ def bootstrap(boot_ip=None, authorized_keys='authorized_keys', static_ip=True):
...
-                'echo """%s""" >> /etc/dhcpcd.conf' %
-                eth_interface.format(**AV))
+        fab.sudo("systemctl stop dhcpcd")
+        fab.sudo("systemctl disable dhcpcd")
+        if configure_ethernet == "yes":
+            eth_config = eth_interface.format(**AV)
+            fab.sudo('echo """%s""" > /etc/network/interfaces.d/%s' %
+                (eth_config, AV['eth_iface']))
+            fab.sudo('echo "source-directory /etc/network/interfaces.d" > /etc/network/interfaces')
+            fab.sudo('echo "nameserver %s" | resolvconf -a %s' %
```
grunskis: Maybe we should uninstall `openresolv` already here and simply do `echo "nameserver 8.8.8.8" > /etc/resolv.conf` here?

Same goes for `dhcpcd`, IMHO we could purge it already here so then we boot in a good state...

deployment/roles/hub-ubuntu/templates/interfaces
```diff
@@ -1,5 +1 @@
-auto {{ploy_wlan0}}
-auto {{ploy_wlan1}}
-
-# Include files from /etc/network/interfaces.d:
 source-directory /etc/network/interfaces.d
```
grunskis: We already create this file from fabfile. Do we need this here?

senic_hub/backend/views/setup_wifi.py
```diff
@@ -31,8 +31,7 @@ class JoinWifiSchema(colander.MappingSchema):
 def scan_wifi_networks(request):
     fs_path = request.registry.settings['wifi_networks_path']
     if os.path.exists(fs_path):
-        networks = json.load(open(fs_path))
-        return list(networks.keys())
+        return json.load(open(fs_path))
```
grunskis: Does `json.load` close the file? If not, we should close it...

senic_hub/backend/views/setup_wifi.py
```diff
@@ -49,34 +48,44 @@ def join_network(request):
...
-    except CalledProcessError:
-        run([
-            'sudo',
-            os.path.join(request.registry.settings['bin_path'], 'enter_wifi_setup'),
-            '-c', request.registry.settings['config_ini_path']
-        ])
-        # TOOD: If we can still respond that probably means wifi wasn't joined,
-        #       or that we were already connected to the same Wi-Fi.
-        #       Study all possible cases to return something helpful if possible.
+        logger.info("Joining network '%s' succeeded" % ssid)
```
grunskis: should be `logger.info("Joining network '%s' succeeded", ssid)`

senic_hub/backend/views/setup_wifi.py
```diff
@@ -49,34 +48,44 @@ def join_network(request):
...
+    except CalledProcessError as e:
+        logger.error("Failed to join network '%s'" % ssid)
```
grunskis: should be `logger.error("Failed to join network '%s'", ssid)`

senic_hub/backend/views/setup_wifi.py
```diff
@@ -49,34 +48,44 @@ def join_network(request):
...
+    ssid = ssid_match.group(1) if ssid_match else None
+    if 'infra_status=connecting' in status:
+        status = 'connecting'
+    elif 'infra_status=connected' in status:
+        status = 'connected'
     else:
-        return dict(ssid=None, status='unavailable')
+        status = 'unavailable'
+    return dict(
+        ssid=ssid if ssid else None,
```
grunskis: Is this `if/else` really needed here?

senic_hub/backend/views/setup_wifi.py
```diff
@@ -49,34 +48,44 @@ def join_network(request):
...
-        return FileResponse(fs_path)
+    try:
+        status = run([
+            'sudo',
+            os.path.join(request.registry.settings['bin_path'], 'wifi_setup'),
+            '-c', request.registry.settings['config_ini_path'],
+            'status'
+        ]).stdout.decode('utf8')
+    except TimeoutExpired:
+        status = ""
```
grunskis: could just return "unavailable"  here w/o checking regexps and stuff

senic_hub/backend/wifi_setup.py
```diff
@@ -0,0 +1,160 @@
...
+@wifi_setup.command(name='join', help="join a given wifi network (requires root privileges)")
+@click.pass_context
+@click.argument('ssid')
+@click.argument('password')
+def wifi_setup_join(ctx, ssid, password):
+    device = ctx.obj['wlan_infra']
+    # Stop wifi scanner and DHCP daemon only if same wlan device is used for adhoc and infrastructure network
+    if device == ctx.obj['wlan_adhoc']:
+        click.echo("Stopping wifi scanner and bringing down ad-hoc network")
+        run(['/usr/bin/supervisorctl', 'stop', 'scan_wifi'])
```
grunskis: Another option would be to use Supervisor RPC interface. See `supervisor.py`...

senic_hub/backend/wifi_setup.py
```diff
@@ -0,0 +1,160 @@
...
+            click.echo("Requesting infrastructure connection state")
+            status = run([
+                os.path.join(ctx.obj['bin_path'], 'wifi_setup'),
+                '-c', ctx.obj['config_ini_path'],
+                'status'
+            ]).stdout.decode()
+            if 'infra_status=connecting' in status:
+                continue
+            connected = 'infra_status=connected' in status
+            break
```
grunskis: These last 4 lines look quite awkward IMO. Why not simply:
```
if 'infra_status=connected' in status:
    break
```

setup.py
```diff
@@ -71,10 +71,8 @@ def find_data(top):
         [paste.app_factory]
         main = senic_hub.backend:main
         [console_scripts]
-        scan_wifi = senic_hub.backend.commands:scan_wifi
-        enter_wifi_setup = senic_hub.backend.commands:enter_wifi_setup
-        join_wifi = senic_hub.backend.commands:join_wifi
-        setup_nuimo = senic_hub.backend.commands:setup_nuimo
```
grunskis: Was removal of `setup_nuimo` accidental?

deployment/fab_pi3.py
larsblumberg: For the records: As discussed, we need to set the gateway as otherwise provisioning the Pi wouldn't work as we need to download files from the internet during provisioning. When a developer uses ethernet to provision/log into Pi, he needs to be aware that IP wouldn't work any longer when the ethernet cable gets unplugged, even when the device is provisioned with wifi.

deployment/fab_pi3.py
larsblumberg: As far as I remember we want the fab file to do the minimal changes necessary to get the provisioning started.

deployment/roles/hub-ubuntu/templates/interfaces
larsblumberg: As far as I understand, the playbook should not rely on the fact which fab file was used to get the provisioning bootstrapped. Sourcing interfaces.d should happen with any machine, no?

senic_hub/backend/views/setup_wifi.py
larsblumberg: Addressed in @e658630

senic_hub/backend/views/setup_wifi.py
larsblumberg: Why is it better than doing string interpolation?

senic_hub/backend/views/setup_wifi.py
larsblumberg: True, but if we placed the early exit here, then we would have to encode the json response object twice: once here in the early exit and then again at the actual return at the end of the function. What do you think?

senic_hub/backend/views/setup_wifi.py
larsblumberg: Yes, it's to not return an empty string if `ssid ==""`. It's indeed hard to read so I added a comment: @9a1d28b

senic_hub/backend/wifi_setup.py
larsblumberg: Thanks! We however also use `run` to control system services through `systemctl`, i.e.:

    run(['/usr/bin/supervisorctl', 'start', 'scan_wifi'])
    run(['/bin/systemctl', 'restart', 'avahi-daemon'])

It looks quiet homogenous to see the same style of starting/stopping services at least for the wifi setup. What do you think? Should also use wrappers for `systemd` then?

senic_hub/backend/wifi_setup.py
larsblumberg: True, looks a little awkward but we also want to break if status is neither `connecting` nor `connected`. I made it a more obvious now: @d505c3e. Is that readable enough?

setup.py
larsblumberg: `setup_nuimo` doesn't exist 😄

senic_hub/backend/wifi_setup.py
tomster: i, too, think that consistency is more important than elegance :)

in this context i'm okay to leave the `supervisorctl` calls in, the actual gain by replacing it with rpc calls is minimal, especially since these aren't called more than a few times in total

senic_hub/backend/wifi_setup.py
```diff
@@ -0,0 +1,161 @@
...
+        click.echo("Failed to join network '%s'" % ssid)
+        exit(1)
+
+
+@click.command(help='scan the wifi interfaces for networks (requires root privileges)')
+@click.option('--config', '-c', required=True, type=click.Path(exists=True), help='app configuration file')
+@click.option('--forever/--no-forever', default=False, help='scan forever (until interrupted)')
+@click.option('--waitsec', default=20, help='How many seconds to wait inbetween scans (only when forever')
+def scan_wifi(config, forever=False, waitsec=20):
+    app = get_app(abspath(config))
```
tomster: why can't this get the config/settings via `pass_context` like the others?

senic_hub/backend/views/setup_wifi.py
larsblumberg: I wanted to point out that we will then have two places where we repeat the response. It's error prove in regard to when we modify how the response looks like. The purpose of the try except block is to retrieve the `status`. If You think in functions that block would only return the status and then be further processed. At this point the response shouldn't be specified. I too like early exits but here I find out cleaner to first retrieve the status and then do further procesing, even if we already now that wifi connection is unavailable.

senic_hub/backend/wifi_setup.py
larsblumberg: I decided to have `scans wifi` not be a sub command of `wifi_setup` although possible. Because this command can run in parallel with the other sub commands. I.e. you can still do scanning while starting the WiFi setup or joining a network.

senic_hub/backend/views/setup_wifi.py
grunskis: Nothing wrong per-se but it's a good style to let logger do the string interpolation only if it's necessary b/c it "knows" when the line is going to be logged and when not.

senic_hub/backend/views/setup_wifi.py
grunskis: I asked b/c the same logic happens 8 lines up if I'm not mistaken and here it's not really needed

senic_hub/backend/wifi_setup.py
grunskis: By using the RPC we could also replace running things in sub-shell and parsing output, like:

```
dhcpd_status = run(
            ['/usr/bin/supervisorctl', 'status', 'dhcpd'],
            stdout=PIPE)
dhcpd_is_running = 'RUNNING' in dhcpd_status.stdout.decode()
```
could be very well be:

```
dhcp_info = supervisor.getProcessInfo('dhcpd')
dhcpd_is_running = dhcp_info['statename'] == 'RUNNING'
```

The gain might be minimal, but the code is much cleaner IMHO

senic_hub/backend/wifi_setup.py
grunskis: My point was that `continue` is not needed b/c you're in an infinite loop anyways, so you could only check for conditions to break out of it.

senic_hub/backend/wifi_setup.py
tomster: good point, except that @larsblumberg now changed dhcpd to be run via systemd and no longer via supervisor :)

senic_hub/backend/wifi_setup.py
tomster: that makes sense

senic_hub/backend/views/setup_wifi.py
tomster: IDNKT, thanks!

senic_hub/backend/wifi_setup.py
larsblumberg: Addressed in @082bbd9

senic_hub/backend/views/setup_wifi.py
larsblumberg: Addressed in f633b09

senic_hub/backend/wifi_setup.py
larsblumberg: Addressed in 4ed3aa4

senic_hub/backend/views/setup_wifi.py
larsblumberg: Addressed in dac47c3

senic_hub/backend/views/setup_wifi.py
larsblumberg: Addressed in dac47c3

senic_hub/backend/views/setup_wifi.py
larsblumberg: At the end are removed the timeout: 940c795

PR #102, opened 2017-04-05
Show error icon on Nuimo when something goes wrong with calling Home Assistant

grunskis:
In the process refactored the the way how websocket communications are done to make this happen. The refactoring was also necessary to be able to better time requests for smoother brightness / volume update that I'll work on next...

### DIFF
--- a/senic_hub/nuimo_app/hass.py
+++ b/senic_hub/nuimo_app/hass.py
@@ -1,10 +1,10 @@
 import json
 import logging
-import time
 
 from collections import defaultdict
 from pprint import pformat
 from threading import Thread
+from time import sleep, time
 
 import websocket
 
@@ -12,13 +12,68 @@ import websocket
 logger = logging.getLogger(__name__)
 
 
+class HomeAssistantConnection(Thread):
+    def __init__(self, url, on_open, on_message, on_close):
+        super().__init__()
+
+        self.websocket_url = "ws://{}/api/websocket".format(url)
+
+        self.on_open_callback = on_open
+        self.on_message_callback = on_message
+        self.on_close_callback = on_close
+
+        self.websocket = None
+
+    @property
+    def connected(self):
+        return self.websocket and self.websocket.sock.connected
+
+    def on_open(self, ws):
+        logger.info("Connected to Home Assistant url: %s", self.websocket_url)
+
+        self.on_open_callback()
+
+    def on_message(self, ws, message):
+        self.on_message_callback(json.loads(message))
+
+    def on_error(self, ws, error):
+        logger.error("Connection failed with: %s", error)
+        self.websocket = None
+
+    def on_close(self, ws):
+        self.websocket = None
+        self.on_close_callback()
+
+    def send(self, message):
+        self.websocket.send(json.dumps(message))
+
+    def run(self):
+        logger.info("Connecting to Home Assistant url: %s...", self.websocket_url)
+
+        self.websocket = websocket.WebSocketApp(
+            self.websocket_url,
+            on_open=self.on_open,
+            on_message=self.on_message,
+            on_error=self.on_error,
+            on_close=self.on_close,
+        )
+        self.websocket.run_forever()
+
+    def stop(self):
+        if self.websocket:
+            self.websocket.close()
+            self.websocket = None
+
+
 class HomeAssistant(Thread):
     """
-    Listens on events & state change notification through HASS WS interface.
+    Listens on events & state change notification through Home
+    Assistant websocket interface and sends commands to services
+    registered with Home Assistant.
 
     """
-    def __init__(self, url, ws_connection=None, on_connect=None, on_disconnect=None):
-        super().__init__(daemon=True)
+    def __init__(self, url, on_connect=None, on_disconnect=None):
+        super().__init__()
 
         self.request_id = 1
         self.event_subscription_id = None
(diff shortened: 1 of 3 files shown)


senic_hub/nuimo_app/hass.py
```diff
@@ -6,13 +6,13 @@
 from pprint import pformat
 from threading import Thread
 
-from websocket import WebSocketConnectionClosedException, create_connection
+import websocket
```
tomster: was this meant to be part of this commit?

senic_hub/nuimo_app/hass.py
```diff
@@ -29,116 +84,106 @@ def __init__(self, url, ws_connection=None, on_connect=None, on_disconnect=None)
...
         self.stopping = False
-        self.response_callbacks = {}  # req_id: callback
 
+        self.result_callbacks = {}  # req_id: callback
         self.state_listeners = defaultdict(list)  # entity_id: [callback1, ...]
 
-    def connect(self, reconnect=False, reconnect_timeout=5):
-        self.connection = None
+    def connect(self):
+        self.connection = HomeAssistantConnection(
```
tomster: does this mean, that the connection instance is now created adhoc, IOW on demand? why was this change necessary or what is the benefit? (just trying to understand)

senic_hub/nuimo_app/hass.py
```diff
@@ -29,116 +84,106 @@ def __init__(self, url, ws_connection=None, on_connect=None, on_disconnect=None)
...
-            if not result:
-                return
-
-            if not (result["id"] == request["id"] and
-                    result["type"] == "result" and
-                    result["success"] is True):
-
-                logger.error(result)
-                return None
+            if self.connection and self.connection.connected:
```
tomster: nice!

senic_hub/nuimo_app/tests/test_hass.py
```diff
@@ -1,17 +1,15 @@
...
-from unittest.mock import MagicMock
 
 from senic_hub.nuimo_app.hass import HomeAssistant
 
 
 class TestHomeAssistant(TestCase):
     def test_find_state_entity_state(self):
-        connection = MagicMock()
-        listener = HomeAssistant("ws://localhost:8123", connection)
+        ha = HomeAssistant("ws://localhost:8123")
```
tomster: does this mean, that test runs require a running HA instance? this commit seems to only remove the mocking without replacing it with something else

senic_hub/nuimo_app/hass.py
tomster: i know it's not a big deal, i'm just a sucker for these little things...

senic_hub/nuimo_app/tests/test_hass.py
grunskis: The connection is not created in constructor anymore so there's no need for the mock. The URL is required thought, but I could probably replace it with some dummy one...

senic_hub/nuimo_app/hass.py
grunskis: Yep, it's legit. The websocket library supports 2 ways of using it - the simple one (it's what I used before) and async one (based on callbacks) which I'm using now

senic_hub/nuimo_app/hass.py
grunskis: Not really. In fact I'm trying to always have connection available if possible (if Home Assistant is up and running).

Does this answer the question? Maybe I misunderstood what you were asking...

senic_hub/nuimo_app/hass.py
grunskis: Didn't get this one. Looks like something is missing in your comment...

senic_hub/nuimo_app/tests/test_hass.py
grunskis: Addressed here 6308663

senic_hub/nuimo_app/hass.py
```diff
@@ -1,24 +1,79 @@
...
+    def __init__(self, url, on_open, on_message, on_close):
+        super().__init__()
+
+        self.ws_url = "ws://{}/api/websocket".format(url)
+
+        self.on_open_callback = on_open
+        self.on_message_callback = on_message
+        self.on_close_callback = on_close
+
+        self.ws = None
```
larsblumberg: I would name it `websocket`

senic_hub/nuimo_app/hass.py
```diff
@@ -1,24 +1,79 @@
...
 
 logger = logging.getLogger(__name__)
 
 
-class HAListener(Thread):
+class HomeAssistantConnection(Thread):
+    def __init__(self, url, on_open, on_message, on_close):
+        super().__init__()
+
+        self.ws_url = "ws://{}/api/websocket".format(url)
```
larsblumberg: I would name it `websocket_url`

senic_hub/nuimo_app/hass.py
```diff
@@ -1,24 +1,79 @@
...
+        self.on_open_callback()
+
+    def on_message(self, ws, message):
+        self.on_message_callback(json.loads(message))
+
+    def on_error(self, ws, error):
+        logger.error("Connection failed with: %s", error)
+        self.ws = None
+
+    def on_close(self, ws):
```
larsblumberg: I would make those methods "private" (prefix with `_`) so the user of this class does not accidentally assign his callbacks to `home_assistant.on_close` instead of `home_assistant.on_close_callback`

senic_hub/nuimo_app/hass.py
```diff
@@ -29,116 +84,106 @@ def __init__(self, url, ws_connection=None, on_connect=None, on_disconnect=None)
...
-            logger.warn("Home Assistant connection lost! Trying to reconnect...")
+        self.next_reconnect_time = time() + self.reconnection_interval
 
-            on_disconnect_listener = self.connection_state_listeners["on_disconnect"]
-            if on_disconnect_listener:
-                on_disconnect_listener()
+    def on_connect(self):
+        on_connect_listener = self.connection_state_listeners["on_connect"]
+        if on_connect_listener:
+            on_connect_listener()
```
larsblumberg: Can be made functional by writing:
```python
(self.connection_state_listeners["on_connect"] or (lambda: None))()
```

senic_hub/nuimo_app/hass.py
```diff
@@ -29,116 +84,106 @@ def __init__(self, url, ws_connection=None, on_connect=None, on_disconnect=None)
...
-        except (ConnectionResetError, WebSocketConnectionClosedException):
-            self.connect(reconnect=True)
+    def on_disconnect(self):
+        if self.connection:
+            self.connection.stop()
+            self.connection = None
+
+        on_disconnect_listener = self.connection_state_listeners["on_disconnect"]
+        if on_disconnect_listener:
+            on_disconnect_listener()
```
larsblumberg: Same functional craziness could be applied here 🐵

senic_hub/nuimo_app/hass.py
```diff
@@ -29,116 +84,106 @@ def __init__(self, url, ws_connection=None, on_connect=None, on_disconnect=None)
...
+            self.url,
+            on_open=self.on_connect,
+            on_message=self.on_message,
+            on_close=self.on_disconnect,
+        )
+        self.connection.start()
 
-        if reconnect:
-            logger.warn("Home Assistant connection lost! Trying to reconnect...")
+        self.next_reconnect_time = time() + self.reconnection_interval
```
larsblumberg: I would only remember when the connection was started and then later check if enough time has passed by to try a reconnection. I find that approach better because the implementation for what a "reconnection time" means is not encoded here but actually in the place where it's needed, namely line 181.

senic_hub/nuimo_app/hass.py
```diff
@@ -29,116 +84,106 @@ def __init__(self, url, ws_connection=None, on_connect=None, on_disconnect=None)
...
-            elif result["type"] == "result":
-                if self.process_response_callbacks(result):
-                    continue
-                else:
-                    logger.debug("Got result w/o callback: %s", result)
-            else:
-                logger.debug("Unknown HA message: %s", result)
+            if not (self.connection and self.connection.connected):
+                if time() >= self.next_reconnect_time:
+                    self.connect()
```
larsblumberg: 1. Given the above comment (remembering start time of connection), I would type here: if time() >= self.connection_time + self.reconnection_interval:

2. Why not combining both if statements into a single if statement with and-joined conditions?

senic_hub/nuimo_app/hass.py
tomster: it was just a comment on my comment :)

senic_hub/nuimo_app/hass.py
grunskis: The callbacks are passed in as parameters to the init method and not assigned directly.

senic_hub/nuimo_app/hass.py
grunskis: I would rather do this:

```
def noop():
    pass

...
    def __init__(self, url, on_connect=noop, on_disconnect=noop):
        super().__init__()

        self.connection_state_listeners = {
            "on_connect": on_connect,
            "on_disconnect": on_disconnect,
        }

...

on_disconnect_listener = self.connection_state_listeners["on_disconnect"]
on_disconnect_listener()
```

But I'm not sure if this is not making the code too smart.

senic_hub/nuimo_app/hass.py
grunskis: Renamed ca99fa1

senic_hub/nuimo_app/hass.py
grunskis: Discussed this IRL with @larsblumberg and decided to leave it as is for now and observe how it behaves...

PR #107, opened 2017-04-06
Frontend tweaks for alpha release

larsblumberg:


deployment/roles/hub-hass/templates/requirements.txt
```diff
@@ -0,0 +1,5 @@
+pip==9.0.1
```
grunskis: This one is actually not necessary here. It's being updated already a few lines up

senic_hub/frontend/src/SetupDevices.js
```diff
@@ -68,9 +68,10 @@ class SetupDevices extends Component {
     fetch('/-/setup/devices/' + device.id + '/authenticate', {method: 'POST'})
       .then((response) => response.json())
       .then((response) => {
-        let i = this.state.devices.findIndex((d) => d.id == device.id)
-        this.state.devices[i].authenticated = response.authenticated
-        this.setState({ devices: this.state.devices })
+        device = this.state.devices.find((d) => d.id === device.id)
```
grunskis: Do we actually need to search for the device object? It's already passed in to the function as `device` parameter

Issue #90, opened 2017-03-24
Frontend is not included in the senic_hub package


larsblumberg:
`/srv/senic_hub/venv/htdocs` is not present after installing `senic_hub` from pypi

tomster:
That's not the intended behaviour. `htdocs` only exists as a destination for creating the pre-built documentation. only the sources are part of the package, not the built artefact, they are part of the devpi publishing process and/or can be built on-demand from the sources.

larsblumberg:
So how do we solve that?

tomster:
why does it need to be solved? we don't want to install pre-generated html docs when installing the egg. at least i don't know why and where to. by default all included files are kept inside the python's `lib/python/site-packages` directory which is not a place where users would look for documentation in my experience.

did you work with another python package that did that and expected this here, too or what is the source of the expectation?

larsblumberg:
When I install the `senic_hub` package (i.e. `pip install senic_hub`) I expect the frontend files to be included. Otherwise the Senic Hub wouldn't completely work after installing the package, right?

tomster:
right, now i understand :)

this indeed has become broken during restructuring. i fixed this now in #110

PR #110, opened 2017-04-10
#90 pip install frontend

tomster:


tomster:
actually, this would be a good opportunity to rename the folder from the generic `htdocs` perhaps to `senic_hub_frontend`?

grunskis:
LGTM! Do the rename as well then ;)

grunskis:
👍

deployment/roles/hub-ubuntu/defaults/main.yml
```diff
@@ -3,5 +3,5 @@ ploy_adhoc_channel: 1
 wwwuser: www-data
 hostname: "{{ploy_hostname}}-{{ansible_machine_id[:4]}}"
 backend_deploy_location: /srv/senic_hub
-frontend_document_root: /srv/senic_hub/venv/htdocs
```
larsblumberg: the `venv` folder is located inside the `senic_hub` folder. Why do you replicate this prefix in `senic-hub-htdocs`?

setup.py
```diff
@@ -35,9 +35,10 @@ def find_data(top):
     package_data={
         name: [
             '.coveragerc',
-            'backend/tests/*.py',
-            'backend/tests/data/*.*',
-            'backend/views/*.*',
+            'senic_hub/backend/tests/*.py',
+            'senic_hub/backend/tests/data/*.*',
+            'senic_hub/backend/views/*.*',
+            'senic_hub/frontend/distribution',
```
larsblumberg: When we run `npm run build` the frontend output is put into the `build` folder, not `distribution`. Can we unify the names?

tomster: unfortunately both requirements are hardcoded by setuptools and npm respectively, the only working solution i found was to rename via the Makefile

deployment/roles/hub-ubuntu/defaults/main.yml
larsblumberg: As discussed orally, `pip install` can be done into any `venv`, it mustn't necessarily be inside our `senic_hub` folder.

setup.py
larsblumberg: Got it, `build` is a reserved name

PR #116, opened 2017-04-11
Try reconnecting to Nuimo even when connection fails

grunskis:
Raising an exception / `sys.exit` wouldn't break out of the main loop, but that's fine for our use case


larsblumberg:
LGTM

PR #123, opened 2017-04-18
Build frontend on travis

tomster:
you can see the output here https://www.travis-ci.org/tomster/nuimo-hub-backend


.travis.yml
```diff
@@ -1,14 +1,22 @@
 git:
   depth: 10000
-sudo: false
+sudo: required
 language: python
+cache:
+  directories:
+    - $HOME/.cache/pip
+    - $HOME/.nvm/
```
grunskis: Is this used by nodejs or npm?

Makefile
```diff
@@ -1,5 +1,6 @@
 # convenience makefile to set up the backend for local development
 python_version = 3
+NODE_BIN = dirname `which node`
```
grunskis: Why this isn't `which npm` instead since that's what we're actually interested in?

travis_after_success.sh
```diff
@@ -4,15 +4,12 @@
...
-echo `git describe --tags`
-echo `git tag`
-
 # build and publish for
 if  [ "$TRAVIS_PULL_REQUEST" = "false" ] ; then
     if [ "$TRAVIS_BRANCH" = "master" ] || [ "$REPO_OWNER" != "getsenic" ]; then
         devpi use "https://pypi.senic.com/$REPO_OWNER/master"
         devpi login $devpi_user --password="$devpi_password"
-        devpi upload --no-vcs --with-docs --formats bdist_wheel
+        make upload NODE_BIN=$HOME/nodejs/node-v6.10.2-linux-x64/bin/
```
grunskis: Why not just overriding `PATH` variable like`make upload PATH=$HOME/nodejs/node-v6.10.2-linux-x64/bin/:$PATH`? In this way you could avoid touching the Makefile...

tomster: good idea!

tomster: meaning: will do, hang on :)

.travis.yml
```diff
@@ -5,7 +5,7 @@ language: python
 cache:
   directories:
     - $HOME/.cache/pip
-    - $HOME/.nvm/
+    - $HOME/.npm/
```
grunskis: Are you sure that NPM installs the libs here? AFAIK they usually end up in `node_modules` directory unless the global mode is turned on.

tomster: true, the global modules aren't particularly interesting in this context. i'll try to cache the local ones instead

PR #124, opened 2017-04-19
Control Sonos & Philips Hue w/o Home Assistant

grunskis:
Controls Sonos & Philips Hue lights directly using their Python SDKs (soco & phue).

To test you would need to rerun the on-boarding and stop Home Assistant.


senic_hub/nuimo_app/__init__.py
```diff
@@ -59,55 +55,32 @@ def __init__(self, ha_api_url, ble_adapter_name, mac_address, components):
...
-
-            # set active component if it's not set already
-            if not self.active_component:
-                self.set_active_component()
+        if self.active_component:
+            self.active_component.stop()
 
-        self.ha.get_state(component.entity_id, [set_state], self.show_error_icon)
+        if self.controller.is_connected():
+            self.controller.disconnect()
```
larsblumberg: You can simply call `disconnect` w/o testing for whether it's connected or not – it should work in both cases.

senic_hub/nuimo_app/__init__.py
```diff
@@ -150,50 +117,23 @@ def process_internal_gesture(self, gesture):
...
-            action = self.active_component.rotation(self.rotation_value)
-            self.rotation_value = 0
-            self.action_in_progress = action
-            return action
-
-    def process_gesture(self, gesture):
-        action = None
+    def process_gesture(self, gesture, delta):
+        if gesture == Gesture.ROTATION:
+            self.active_component.rotation(delta)
```
larsblumberg: `rotation()`: To me nouns are reserved for types. I would name this one `on_rotation` or `rotated()`.

senic_hub/nuimo_app/__init__.py
```diff
@@ -150,50 +117,23 @@ def process_internal_gesture(self, gesture):
...
 
         if gesture == Gesture.BUTTON_PRESS:
-            action = self.active_component.button_press()
+            self.active_component.button_press()
```
larsblumberg: `on_button_press` or `button_pressed`

(following gestures handlers I would rename analogously)

senic_hub/nuimo_app/__init__.py
```diff
@@ -226,20 +166,14 @@ def set_active_component(self, component=None):
...
-            self.ha.register_state_listener(self.active_component.entity_id, self.state_changed)
-
-    def state_changed(self, state):
-        """
-        Gets called whenever state changes in any device within
-        currently active component group.
-
-        """
+            self.component_thread = Thread(target=self.active_component.run)
+            self.component_thread.start()
```
larsblumberg: I'd like to understand the need for a thread per component. I see a `active_component.stop()` call here and would also expect a `active_component.start()` call. Depending on how the component thread is used (I'm still in progress of understanding), the `start()` method could create the thread and return it.

senic_hub/nuimo_app/__init__.py
```diff
@@ -248,18 +182,10 @@ def show_active_component(self):
         else:
             icon = icons.ERROR
```
larsblumberg: It would be good if we stick with the terminology "matrix" and not also introduce the word "icon". This line then would read: `matrix = matrices.ERROR`.

senic_hub/nuimo_app/__init__.py
```diff
@@ -248,18 +182,10 @@ def show_active_component(self):
...
-        self.controller.display_matrix(
-            matrix_config.matrix,
-            fading=matrix_config.fading,
-            ignore_duplicates=matrix_config.ignore_duplicates,
-        )
-
-    def quit(self):
-        if self.controller.is_connected():
-            self.controller.disconnect()
+    def display_matrix(self, icon, **kwargs):
```
larsblumberg: Same here: `icon` -> `matrix`

senic_hub/nuimo_app/__main__.py
```diff
@@ -56,11 +58,14 @@ def read_config(config_file_path):
...
+
     for cid in component_config.sections():
         cfg = component_config[cid]
-        component_class = getattr(components, cfg["component"])
-        component = component_class(cfg["name"], cfg["entity_id"])
-        instances.append(component)
+        module_name = module_name_format.format(cfg.pop('component'))
+        logger.info("Importing module %s", module_name)
+        component_module = import_module(module_name)
+        instances.append(component_module.Component(cfg))
```
larsblumberg: Shall we have a comment for all prior lines? From reading them I have no immediate clue what's the requirement behind them.

senic_hub/nuimo_app/components/philips_hue.py
```diff
@@ -0,0 +1,163 @@
...
+
+
+logger = logging.getLogger(__name__)
+
+
+class Component(BaseComponent):
+    ICON = icons.LIGHT_BULB
+
+    # don't send updates to the bridge more often than UPDATE_INTARVAL
+    UPDATE_INTERVAL = 0.1  # seconds
```
larsblumberg: This update interval is for request for single lights. I'm wondering if we are also controlling groups? They need an interval between two group requests of 1 second.

senic_hub/nuimo_app/components/philips_hue.py
```diff
@@ -0,0 +1,163 @@
...
+
+        self.is_on = None
+        self.brightness = None
+        self.delta = 0
+        self.update_state()
+
+        # seed random nr generator (used to get random color value)
+        seed()
+
+    def get_or_create_entity(self):
```
larsblumberg: What does `entity` stand for?

senic_hub/nuimo_app/components/philips_hue.py
```diff
@@ -0,0 +1,163 @@
...
+
+        return int(group_id)
+
+    def create_light_group(self, lights):
+        response = self.device.create_group(SENIC_HUB_GROUP_NAME, list(lights.keys()))
+        return response[0]['success']['id']
+
+    def update_state(self):
+        try:
+            settings = self.device.get_light(self.entity_id)
```
larsblumberg: Rename `settings` -> `light`? Then it reads:
`light = self.device.get_light(...)`

senic_hub/nuimo_app/components/philips_hue.py
```diff
@@ -0,0 +1,163 @@
...
+            logger.exception()
+
+    def button_press(self):
+        on = not self.is_on
+        if on and self.brightness:
+            self.set_light(on=on, bri=self.brightness)
+        else:
+            self.set_light(on=on)
+
+    def set_light(self, **attributes):
```
larsblumberg: `set_light_attributes`? Because it's not setting a light but its attributes.
Why do we hide which actual parameters can be passed? I believe we should be explicit here about which parameters can be set. I can see that you are forwarding these to `device.set_light()` but we shouldn't expose the same obscurity that the phue library uses.

senic_hub/nuimo_app/components/philips_hue.py
```diff
@@ -0,0 +1,163 @@
...
+
+                elif attribute == 'on':
+                    self.is_on = value
+
+                    if self.is_on:
+                        icon = icons.LIGHT_ON
+                    else:
+                        icon = icons.LIGHT_OFF
+
+                    self.nuimo.display_matrix(icon, fading=True, ignore_duplicates=True)
```
larsblumberg: `fading` and `ignore_duplicates` should be both `False`. Because we want to quick button presses to flicker the LED. Generally, all discrete actions such as light on/off, next song, next color, etc., have fading=False and ignore_duplicates=False. Continuous value changes however have them set to True.

senic_hub/nuimo_app/components/sonos.py
```diff
@@ -0,0 +1,157 @@
...
+    STATE_STOPPED = 'STOPPED'
+
+    # how many seconds to wait after sending last command before
+    # receiving device state change events
+    EVENT_IDLE_INTERVAL = 2  # seconds
+
+    def __init__(self, config):
+        super().__init__(config)
+
+        self.device = SoCo(config['ip_address'])
```
larsblumberg: I would name the device `sonos_player`, `sonos_controller` or `soco`. `device` is too abstract, we can be concrete here.

senic_hub/nuimo_app/components/philips_hue.py
```diff
@@ -0,0 +1,163 @@
...
+
+    # how often to sync with the bridge to detect changes made by external apps
+    SYNC_INTERVAL = 5  # seconds
+
+    def __init__(self, config):
+        super().__init__(config)
+
+        self.encoder = EncoderRing(1, 254)
+
+        self.device = Bridge(config['ip_address'], config['username'])
```
larsblumberg: I would name the device `bridge`. `device` is too abstract, we can be concrete here.

senic_hub/nuimo_app/components/sonos.py
```diff
@@ -0,0 +1,157 @@
...
+        self.subscribe_to_events()
+
+        self.state = None
+        self.volume = None
+        self.update_state()
+
+        self.nuimo = None
+
+        self.stopping = False
+        self.last_command_time = time()
```
larsblumberg: `last_request_time` since we are making HTTP requests?

senic_hub/nuimo_app/components/__init__.py
```diff
@@ -0,0 +1,28 @@
...
+
+    def __init__(self, min_value, max_value):
+        self.min_value = min_value
+        self.max_value = max_value
+
+    def points_to_value(self, points):
+        return points / self.NUM_POINTS * self.max_value
+
+    def normalize_value(self, value):
+        return min(max(value, self.min_value), self.max_value)
```
larsblumberg: min(max()) is not called normalizing (normalizing in the simplest case is transforming a value from one scale to another scale), a common name for min(max()) is: `clamp_value()`

senic_hub/nuimo_app/__init__.py
larsblumberg: Ok, got the need for threads now. It's because HTTP requests are made synchronously.

senic_hub/nuimo_app/components/philips_hue.py
larsblumberg: Get it now – it's either a light or a group. Can we be more explicit here?

senic_hub/nuimo_app/components/philips_hue.py
grunskis: Good point! Until now I was only testing with a single light or a group with a single light! I'll include proper group handling in a separate PR if that's fine with you?

senic_hub/nuimo_app/components/philips_hue.py
grunskis: Will take care of this in the next PRs when I'll add group support.

grunskis:
@larsblumberg Addressed all the feedback. Please take a look...

PR #131, opened 2017-04-27
Control more than one Philips Hue light

grunskis:
1) During on-boarding all the available lights from an authenticated bridge would be added to the device configuration file and the propagated to the Nuimo app config.

2) On startup, Nuimo app would either create a group of lights (if # of lights > 10) and control that or otherwise control each light separately.

Also adding the Philips Hue bridge username to the device configuration file for convenience.

### DIFF
--- a/senic_hub/backend/views/setup_devices.py
+++ b/senic_hub/backend/views/setup_devices.py
@@ -102,7 +102,7 @@ def devices_authenticate_view(request):
     with open(phue_bridge_config, "w") as f:
         json.dump(config, f)
 
-    update_device(device, request.registry.settings)
+    update_device(device, request.registry.settings, username)
 
     return {"id": device_id, "authenticated": authenticated}
 
@@ -152,10 +152,17 @@ def get_device(device_list_path, device_id):
     return device
 
 
-def update_device(device, settings):
+def update_device(device, settings, username):
     devices_path = settings["devices_path"]
     with open(devices_path, "r") as f:
         devices = json.loads(f.read())
+
+    device["extra"]["username"] = username
+
+    if device['authenticated'] and username:
+        bridge = PhilipsHueBridgeApiClient(device["ip"], username)
+
+        device['extra']['lights'] = bridge.get_lights()
 
     device_index = [i for (i, d) in enumerate(devices) if d["id"] == device["id"]].pop()
 
--- a/senic_hub/nuimo_app/components/philips_hue.py
+++ b/senic_hub/nuimo_app/components/philips_hue.py
@@ -1,188 +1,350 @@
 import logging
 
+from pprint import pformat
 from random import random, seed
 from time import sleep, time
 
-from phue import Bridge, PhueRequestTimeout
+from phue import Bridge
 
-from . import BaseComponent, EncoderRing
+from . import BaseComponent, clamp_value, normalize_delta
 
 from .. import matrices
 
 
 COLOR_WHITE_XY = (0.32, 0.336)
-SENIC_HUB_GROUP_NAME = "Senic hub demo"
 
 
 logger = logging.getLogger(__name__)
 
 
+class HueBase:
+    @property
+    def sync_interval(self):
+        """
+        How often to sync with the bridge to detect changes made by
+        external apps.
+        """
+        return 5  # seconds
+
+    def parse_responses(self, responses, request_attributes):
+        errors = [x['error'] for x in responses[0] if 'error' in x]
+        if errors:
+            return {'errors': errors}
+
+        response = self.merge_success_responses(responses)
+        logger.debug("response: %s", response)
+
+        self.update_state_from_response(response)
+        return response
+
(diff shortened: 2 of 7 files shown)


larsblumberg:
As we just discussed, we want to send the ON/OFF command always to a group to have all lights (if more than one) switch on/off at the very same time

senic_hub/nuimo_app/components/philips_hue.py
```diff
@@ -1,188 +1,351 @@
...
-    # don't send updates to the bridge more often than UPDATE_INTARVAL
-    UPDATE_INTERVAL = 0.1  # seconds
+    def parse_responses(self, responses, request_attributes):
+        errors = [x['error'] for x in responses[0] if 'error' in x]
+        if errors:
+            return {'errors': errors}
 
-    # how often to sync with the bridge to detect changes made by external apps
-    SYNC_INTERVAL = 5  # seconds
+        merged = self.merge_success_responses(responses)
```
larsblumberg: `merged` reads like a boolean variabe to name. Since we don't annotate with types, we should use expressive names that suggest their type or at least don't suggest misleading types. Here maybe: `merged_response`?

senic_hub/nuimo_app/components/philips_hue.py
```diff
@@ -1,188 +1,351 @@
...
-        # TODO handle cases when light becomes available later at some point
-        lights = {k: v for k, v in self.bridge.get_light().items() if v['state']['reachable']}
-        if not lights:
-            return None
-        elif len(lights) == 1:
-            return int(list(lights.keys())[0])
+        return 60  # seconds
+
+
+class Lights(HueBase):
```
larsblumberg: I wouldn't use plural name for classes, here I would use `LightCollection` or `LightSet`.

senic_hub/nuimo_app/components/philips_hue.py
```diff
@@ -1,188 +1,351 @@
...
+    def update_state(self):
+        """
+        Get current state of all the lights from the bridge
+        """
+        response = self.bridge.get_light()
+        self._state = {k: v['state'] for k, v in response.items() if k in self.light_ids}
+
+        logger.debug("state: %s", pformat(self._state))
+
+        self._on = any(s['on'] for s in self._state.values())
```
larsblumberg: Philips Hue bridge considers a set of lights to be off if at least one if their lights is off. This would be contrary to the formula here where you assume all lights on if at least one is on. This said, it should probably be:

    self._on = all(s['on'] for s in self._state.values())

senic_hub/nuimo_app/components/philips_hue.py
```diff
@@ -1,188 +1,351 @@
...
+        if group_light_ids != set(self.light_ids):
+            return self.update_group(int(group_id))
 
-        group_id = int(group_id)
+        return int(group_id)
 
-        if group_lights != all_lights:
-            return self.update_light_group(group_id, lights)
+    def create_group(self):
+        responses = self.bridge.create_group(self.GROUP_NAME, self.light_ids)
```
larsblumberg: Is `create_group` returning multiple responses or a single response? The variable name `responses` suggests that we have multiple responses.

senic_hub/backend/views/setup_devices.py
```diff
@@ -152,11 +152,18 @@ def get_device(device_list_path, device_id):
     return device
 
 
-def update_device(device, settings):
+def update_device(device, settings, username):
     devices_path = settings["devices_path"]
     with open(devices_path, "r") as f:
         devices = json.loads(f.read())
 
+    device["extra"]["username"] = username
```
grunskis: Ok. I'm still getting used to single quotes :)

senic_hub/nuimo_app/components/philips_hue.py
grunskis: Yep, it returns list of responses even though there's usually one item in the list. It's super annoying...

senic_hub/nuimo_app/components/philips_hue.py
grunskis: Addressed: 47994b8

senic_hub/nuimo_app/components/philips_hue.py
grunskis: Thanks! Addressed ba8649b

senic_hub/nuimo_app/components/philips_hue.py
grunskis: Addressed: 4b201f6

grunskis:
I addressed your feedback. Please check...

I also tried to improve the value normalization/clamping 1afb428 Please take a look and give feedback...

The ON/OFF logic with multiple lights I'll address in a separate PR.

grunskis:
Thanks for suggestion. I implemented a version of this here c9fa9eb IMO is much cleaner now...

PR #133, opened 2017-05-08
Add Sonos component that uses Home Assistant's WebSocket API

larsblumberg:


senic_hub/nuimo_app/components/sonos_ha.py
```diff
@@ -0,0 +1,96 @@
...
+            volume = state.get('attributes', {}).get('volume_level', None)
+        else:
+            volume = None
+        if volume is not None:
+            # Only overwrite our volume if we haven't recently modified it ourselves
+            # This avoids synching with an old volume while the user is still turning the wheel
+            if time() - self.last_volume_request_time > 5.0:
+                self.volume = volume
+        else:
+            self.volume = None
```
grunskis: We get here when `self.volume` is already `None`. IMO the logic in this method could be simplified a bit...

senic_hub/nuimo_app/components/__init__.py
```diff
@@ -32,3 +32,65 @@ def start(self):
...
+        self._ha.start()
+        self._ha.register_state_listener(self._ha_entity_id, self._ha_state_changed)
+
+    def stop(self):
+        super().stop()
+        self._ha.stop()
+        self._ha.unregister_state_listener(self._ha_entity_id)
+
+    def update_from_ha_state(self, state):
+        self.is_on = state.get('state', None) != 'off'
```
grunskis: IMHO this reads better `self.is_on = state.get('state', 'off') == 'on'`

senic_hub/nuimo_app/components/sonos_ha.py
```diff
@@ -0,0 +1,96 @@
...
+        self.is_volume_request_pending = False
+        self.last_volume_request_time = 0
+        self.should_send_volume_again = False
+
+    def run(self):
+        super().run()
+
+        # TODO: Do we need this run loop?
+        while not self.stopping:
+            pass
```
grunskis: I think we need to sleep here at least bit. Otherwise I think we're not giving much CPU time to the other threads. Not exactly sure how this works in Python to be honest...

senic_hub/nuimo_app/components/__init__.py
larsblumberg: Yes, your suggestion looks cleaner, I will however use

    self.is_on = state.get('state', 'on') != 'off'

as nearly always power `state` is not part of the update state. So far I've seen it only when it powered off.

senic_hub/nuimo_app/components/__init__.py
```diff
@@ -32,3 +32,65 @@ def start(self):
...
+        self._ha.call_service(self._ha_domain, service, data, on_success or _on_success, on_error or _on_error)
+
+    def _ha_connected(self):
+        def on_state_retrieved(state):
+            logger.debug("on_state_received")
+            if self.stopping:
+                return
+            self._ha_state_changed(state)
+
+        self._ha.get_state(self._ha_entity_id, on_state_retrieved, self._ha_get_state_failed)
```
larsblumberg: Good catch – will make the other method local as well

larsblumberg: f45c637

senic_hub/nuimo_app/matrices.py
```diff
@@ -132,13 +135,14 @@
     "         "
 
 
-def light_bar(max_value, value):
+def volume_bar(progress):
```
larsblumberg: Yes, volume isn't suitable for all cases, I however find this name less ambiguous than "light bar" since "light" has multiple meanings in English. And as Nuimo's by far biggest use case is volume control, I renamed it to "volume bar" as long as we find a better name.

senic_hub/nuimo_app/matrices.py
```diff
@@ -132,13 +135,14 @@
...
     """
     Generates a light bar matrix to display volume / brightness level.
 
+    :param progress: value between 0..1
```
larsblumberg: Before we were using values from 0..1800 to describe rotation values on a scale of with a number of 1800 ticks for a full rotation. Code consuming this number becomes easier to read and maintain if we normalize the rotation value from 0..1800 -> 0..1 already here. Additionally we map volume and brightness change from zero to max to a full rotation of Nuimo, hereby a value of "1" is very suitable as it indicates one full rotation.

senic_hub/nuimo_app/matrices.py
```diff
@@ -132,13 +135,14 @@
...
     """
     dots = list(" " * 81)
-    num_dots = int(value / max_value * 9)
+    num_dots = ceil(round(progress, 3) * 9)
```
larsblumberg: Previously we mapped the progress to the number of dots differently, we were showing "no dots" much earlier, i.e. all progress below 11.11...% was mapped to "no dots". For the UX it's better if we show no dots only if volume has reached 0%, anything above 0% should show at least one dot. Furthermore, 9 dots were only shown when 100% volume was reached, but we can show 9 dots also when we're slightly below maximum. This said, the new formula maps as following:

    [0%] -> 0 dots
    (0, 11.11%] -> 1 dot
    (11.11%, 22.22%] -> 2 dots
    ...

senic_hub/nuimo_app/components/sonos_ha.py
```diff
@@ -0,0 +1,96 @@
...
+        super().__init__("media_player", config)
+
+        self.playback_state = None
+        self.volume = None
+        self.is_volume_request_pending = False
+        self.last_volume_request_time = 0
+        self.should_send_volume_again = False
+
+    def run(self):
+        super().run()
```
larsblumberg: Good point! I moved initializing the `stopping` variable to `BaseComponent`'s `start`, that's the right place as `stop` consequently resets it again back to `False`: 37f1de8

senic_hub/nuimo_app/components/sonos_ha.py
```diff
@@ -0,0 +1,96 @@
...
+
+        # TODO: Do we need this run loop?
```
larsblumberg: What would be the issue if the thread wasn't running any longer?

senic_hub/nuimo_app/components/sonos_ha.py
larsblumberg: Yup, now done here: b2730e6

senic_hub/nuimo_app/components/sonos_ha.py
```diff
@@ -0,0 +1,96 @@
...
+        self.nuimo.display_matrix(matrices.volume_bar(self.volume), fading=True, ignore_duplicates=True)
+
+    def on_button_press(self):
+        # TODO: When triggering play state change, we won't know very soon which new state the player is in.
+        #       This said, when the user continuosly presses the button we can't quickly switch playback state yet.
+        if not self.is_on:
+            self.nuimo.display_matrix(matrices.ERROR)
+        elif self.playback_state == 'playing':
+            self.nuimo.display_matrix(matrices.PAUSE)
+            self.call_ha_service("turn_off", {})
```
larsblumberg: Nice, done here: 6a95a43

senic_hub/nuimo_app/components/sonos_ha.py
```diff
@@ -0,0 +1,96 @@
...
+        logger.debug("Updated Sonos state: %s", pformat(state))
+        self.playback_state = state.get('state', None)
+        if self.is_on:
+            volume = state.get('attributes', {}).get('volume_level', None)
+        else:
+            volume = None
+        if volume is not None:
+            # Only overwrite our volume if we haven't recently modified it ourselves
+            # This avoids synching with an old volume while the user is still turning the wheel
+            if time() - self.last_volume_request_time > 5.0:
```
larsblumberg: Done: 35de4b1

senic_hub/nuimo_app/components/sonos_ha.py
larsblumberg: I think you confused `self.volume` with `volume`. As both variables can be easily confused, I renamed the local variable `volume` to `received_volume`: 5fddd10

senic_hub/nuimo_app/matrices.py
```diff
@@ -132,13 +135,14 @@
     "         "
 
 
-def light_bar(max_value, value):
```
larsblumberg: I would like to standardize all methods that are generic and can be used from different components to use generic values. For progress (as here) and Nuimo rotations I want to use the range 0..1 as it's used by most SDKs/libraries as well.

senic_hub/nuimo_app/components/__init__.py
larsblumberg: This said, power state needs to be considers "on" by default if not specified in the state dictionary (as in most of the cases).

senic_hub/nuimo_app/components/__init__.py
```diff
@@ -32,3 +32,65 @@ def start(self):
...
+        self.is_on = False
+
+        # TODO: Parametrize HA's address?
+        self._ha_address = "localhost:8123"
+        # TODO: Read entity id from config
+        self._ha_entity_id = "media_player.office"
+        self._ha_domain = ha_domain
+
+    def run(self):
+        # TODO: `stopping` should be set by `BaseComponent`'s `run()`
```
larsblumberg: Nice – here's an implementation from that we can copy-and-paste: http://stackoverflow.com/a/325528/543875

senic_hub/nuimo_app/components/__init__.py
```diff
@@ -32,3 +32,65 @@ def start(self):
...
+
+        self.is_on = False
+
+        # TODO: Parametrize HA's address?
+        self._ha_address = "localhost:8123"
+        # TODO: Read entity id from config
+        self._ha_entity_id = "media_player.office"
+        self._ha_domain = ha_domain
+
+    def run(self):
```
larsblumberg: Agreed – I moved the initialization into the `start` method: 37f1de8

senic_hub/nuimo_app/components/__init__.py
```diff
@@ -32,3 +32,65 @@ def start(self):
...
+
+
+class HomeAssistantComponent(BaseComponent):
+    def __init__(self, ha_domain, config):
+        super().__init__(config)
+
+        self.is_on = False
+
+        # TODO: Parametrize HA's address?
+        self._ha_address = "localhost:8123"
```
larsblumberg: We just had a great discussion in person :) I now came to the conclusion that I don't consider underscores as a programming style: You can either care about thinking what's "private" and what not and if you do care, you can "annotate" those vars as such. This said, if I find a file that whose author didn't mark some variables as "not to be used in subclasses", IMHO I don't break the coding style be thinking about what's private. But yeah, this is to some extend personal taste if you want to think about this "private" thing or not. Let's continue this discussion in person also with @tomster.

senic_hub/nuimo_app/__init__.py
```diff
@@ -114,7 +114,7 @@ def process_internal_gesture(self, gesture):
 
     def process_gesture(self, gesture, delta):
         if gesture == Gesture.ROTATION:
-            self.active_component.on_rotation(delta)
+            self.active_component.on_rotation(delta / 1800)
```
larsblumberg: This should be the only place ever where we reference the number of rotation ticks for a full rotation. Thus I want to avoid a const placed farer away from the only use. I now documented this magic number: ba49f97

senic_hub/nuimo_app/components/sonos_ha.py
grunskis: Good question :) Not sure really, but I assume that state update callbacks won't be called etc, but to know for sure it's better to try...

senic_hub/nuimo_app/matrices.py
grunskis: We discussed IRL, but just for the record I'll add my view here. IMO saying that "light" has many meanings is not a good reason (volume has at least several meaning too). As you mentioned IRL there's another reason why calling this a volume bar makes sense - Nuimo originally used to be a volume controller and it made sense to call this a volume bar...

senic_hub/nuimo_app/matrices.py
grunskis: Thanks! I was not aware of this algorithm.

Just out of curiosity - why calling the parameter `progress` and not `volume`?

larsblumberg: Good point. What if we rename the method to `progress_bar(progress)`?

PR #137, opened 2017-05-11
Change brightness when rotating nuimo’s wheel

larsblumberg:


larsblumberg:
Travis is failing because of missing tests for previous PR #136, fixed in #138

PR #141, opened 2017-05-17
Add condition to update_from_ha_state()

commento:
Related to a wrong behavior with lightify component. After the light
was turned on the nuimo could not rotate left. After this correction
the use case works better. To be tested if this correction does not
introduce drawback with other lights.

Error Logs:
“
2017-05-16 15:53:50,605 DEBUG nuimo_app.components.light received_brightness: 254, Brightness: None
2017-05-16 15:53:50,606 DEBUG nuimo_app.components.light if received_brightness is not None:
2017-05-16 15:53:50,607 DEBUG nuimo_app.components.light Is On: True, Brightness: None
2017-05-16 15:53:50,657 DEBUG nuimo_app Ignoring gesture event: Gesture.BUTTON_RELEASE
“


senic_hub/nuimo_app/components/light.py
```diff
@@ -30,7 +30,7 @@ def update_from_ha_state(self, state):
         if received_brightness is not None:
             # Only overwrite our brightness if we haven't recently modified it ourselves
             # This avoids synching with an old brightness while the user is still turning the wheel
-            if time() - self.last_brightness_request_time > Component.BRIGHTNESS_CHANGE_RESPONSE_TIMEOUT:
+            if self.brightness == None or time() - self.last_brightness_request_time > Component.BRIGHTNESS_CHANGE_RESPONSE_TIMEOUT:
```
grunskis: The correct way to check if a variable is `None` in python is `if self.brightness is None`

commento: Thanks, I fixed the pep8 error with commit 059e8b4
Now all the checks are passed

larsblumberg:
👍 Looks good to me

PR #143, opened 2017-05-17
Add API for adding components to Nuimo

larsblumberg:


senic_hub/backend/tests/test_nuimo_components.py
```diff
@@ -0,0 +1,136 @@
...
+def temporary_nuimo_app_config_file(settings):
+    """don't run actual external commands during these tests
+    """
+    with NamedTemporaryFile('w+', delete=False) as f:
+        temp_file_name = f.name
+        with open(settings['nuimo_app_config_path']) as config_file:
+            f.write(config_file.read())
+    settings['nuimo_app_config_path'] = f.name
+    yield temporary_nuimo_app_config_file
+    remove(temp_file_name)
```
grunskis: Maybe we can now simplify this and allow `NamedTemproraryFile` to delete the file?

senic_hub/backend/views/nuimo_components.py
```diff
@@ -0,0 +1,114 @@
+import colander
+import logging
+import os.path
+from .setup_devices import get_device
+from ..config import path
```
grunskis: Would be great to order always order imports in this order:

1) imports from python stdlib
2) 3rd party lib imports
3) relative imports from our project

Then in each section we can sort them alphabetically

senic_hub/backend/views/nuimo_components.py
```diff
@@ -0,0 +1,114 @@
...
+    return component
+
+
+@nuimo_component_service.delete()
+def delete_nuimo_component_view(request):
+    id = request.matchdict["component_id"]
+    with open(request.registry.settings['nuimo_app_config_path'], 'r+') as f:
+        config = ConfigParser()
+        config.read_file(f)
+        if id not in config.sections():
```
grunskis: `id` is a built-in name, it's better to use some other name like `section_id`

PR #146, opened 2017-05-19
Device selection

grunskis:
Shows list of configured components after the onboarding and allows to add new components. There's no API to get devices by type so currently you can only add new Philips Hue components (there's a hardcoded deviceId)


senic_hub/setup_app/components/DeviceSelection.js
```diff
@@ -0,0 +1,36 @@
...
+    this.state = {
+      component: props.navigation.state.params,
+    }
+  }
+
+  render() {
+    return (
+      <List>
+        <FlatList
+          data={this.state.component.selected_devices}
```
larsblumberg: Is `selected_devices` coming from our backend's API or why do we use an underscore here?

senic_hub/setup_app/components/SelectComponentDevices.js
```diff
@@ -0,0 +1,53 @@
...
+        'Content-Type': 'application/json',
+      },
+      body: body,
+    }
+    console.log('sending body', body, 'params', params)
+    fetch(API_URL + '/-/nuimos/0/components', params)
+                   .then((response) => {
+                     console.log('response', response)
+                     navigation.navigate('NuimoComponents')
+                   })
```
larsblumberg: This indentation got a little too much attention here ;)

senic_hub/setup_app/components/SetupDevices.js
```diff
@@ -63,27 +67,30 @@ export default class SetupDevices extends Component {
     this.pollDevices()
   }
 
-  componentWillUnmount() {
+  clearTimeouts() {
```
larsblumberg: Why are timeouts not cleared any longer in `componentWillUnmount`? I.e. why was this change necessary?

grunskis: The screen is no longer unmountented when it's part of the navigation. Not sure why they don't do that... Keeping in the `componentWillUnmount` would hurt though. What do you think?

senic_hub/setup_app/components/SelectComponentDevices.js
grunskis: Yep. Still struggling to teach Emacs how to understand ES6+JSX... Will update...

senic_hub/setup_app/components/DeviceSelection.js
grunskis: Yep, coming from the API

senic_hub/setup_app/components/SelectComponentDevices.js
grunskis: Fixed in cf28853

senic_hub/setup_app/components/SetupDevices.js
larsblumberg: For the records: We decided to address that with a different PR, using the WIX navigation component you suggested

PR #148, opened 2017-05-21
Add generic devices API endpoint; add endpoints for getting and modifying nuimo app components

larsblumberg:
- Add a new endpoint `/devices` that returns a flat list of all discovered devices, including philips hue lights
- This endpoint will co-exist for a while with the `/setup/devices` endpoint for now to avoid refactoring too much code until we take time to migrate the device onboarding UI to use the new endpoint
- Add endpoint for getting the selected devices of a nuimo app component
- Add endpoint for modifying the selected devices of a nuimo app component
- Unify `device_id` and `lights` in `nuimo_app.cfg` into one single attribute `device_ids` as in the future also multiple speakers (and other devices) will be selectable per component
- Philips hue lights get their bridge's device ID + suffix `-light-<ID>` to properly map bridges to its lights


senic_hub/backend/tests/test_devices.py
```diff
@@ -0,0 +1,72 @@
...
+            "id": "ph2",
+            "type": "philips_hue",
+            "ip": "127.0.0.2",
+            "virtual": True,
+            "authenticationRequired": True,
+            "authenticated": True,
+        },
+        {
+            "id": "ph2-light-4",
+            "type": "philips_hue",
```
grunskis: Maybe we should have a different type for the lamps or the bridge?

senic_hub/backend/tests/test_nuimo_components.py
```diff
@@ -56,40 +55,33 @@ def temporary_nuimo_app_config_file(settings):
...
+        'type': 'sonos'},
         status=200
     )
 
     config = ConfigParser()
     config.read(settings['nuimo_app_config_path'])
     assert len(config.sections()) == 4
     last_section = dict(config.items(config.sections()[-1]))
-    assert last_section['device_id'] == 's1'
+    assert last_section['device_ids'] == ', '.join(['s1'])
```
grunskis: There's only one item, so could be written as `'s1'`

senic_hub/backend/views/devices.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+def expand_devices(raw_device):
+    """
+    Some devices contain sub-devices, i.e. a Philips Hue bridge contains lights.
+    This method expands such devices and returns the device itself and its
+    contained devices as a list.
+
+    """
+    device = {k: raw_device[k] for k in raw_device.keys() & {
```
grunskis: What is `&` for?

senic_hub/backend/views/nuimo_components.py
```diff
@@ -94,20 +120,55 @@ def create_component(device):
...
     }
     if component_type in ['philips_hue', 'sonos']:
         component['ip_address'] = device['ip']
     if component_type == 'philips_hue':
         component['username'] = device['extra']['username']
-        component['lights'] = ", ".join(sorted(list(device['extra']['lights'])))
+        light_ids = sorted(list(device['extra']['lights']))
+        light_ids = ['%s-light-%s' % (device['id'], i) for i in light_ids]
+        component['device_ids'] = light_ids
+    if component_type == 'light' or component_type == 'media_player':
```
grunskis: Maybe we could introduce a list of home assistant types and then do `if component_type in HOME_ASSISTANT_COMPONENT_TYPES`

senic_hub/backend/tests/test_nuimo_components.py
larsblumberg: As per the description of this PR: we are now storing a list of selected devices per component. So this a generic attribute representing a list of devices that contains one ore devices

senic_hub/backend/views/devices.py
larsblumberg: It selects only items that are common in both sets. This is to make sure that we only copy valid device attributes. I however found a cleaner way to express this logic and fixed it here: 642e34d

PR #152, opened 2017-05-23
Nanopi bootstrap on senicOS

tomster:
not quite finished, but this already installs our hub onto the current dev-version of senicOS ;)

TODO: adjust the homeassistant role, too


deployment/roles/hub-backend/defaults/main.yml
```diff
@@ -2,3 +2,5 @@
 python_interpreter: python3
 backend_deploy_location: /srv/senic_hub
 backend_data_location: /srv/senic_hub/data
+hass_data_location: /srv/homeassistant/data
+locale: en_GB.UTF-8
```
grunskis: Out of curiosity - why we didn't go with `en_US` instead?

deployment/roles/hub-backend/tasks/main.yml
```diff
@@ -65,11 +65,17 @@
     group: "{{run_user}}"
     mode: "0644"
     content: |
+      asn1crypto==0.22.0
+      cffi==1.8.3
       click==6.6
       colander==1.3.1
       cornice==1.2.1
+      cryptoyaml==0.2.0
+      cryptography==1.4
```
grunskis: In case these are too old I've build wheels for newer versions of cryptography & cffi. They are in py pypi index: https://pypi.senic.com/grunskis/master

deployment/roles/hub-backend/defaults/main.yml
tomster: it was the default picked by our french subcontractor :)

and instead of rebuilding the os i opted for adapting it and making it configurable

deployment/roles/hub-backend/tasks/main.yml
tomster: i was thinking about making wheels, i will try if they can be installed, but since ultimately we will be rolling our own recipes for each python package i didn't see the urgency to try building wheels.

we should try it out, though at some point even if just to learn if and how it works cross platform (especially cryptography can be picky in that regard

grunskis: I did build them, so I already have experience with building. That being said, I think it's better to upgrade and use BitBake recipes for this.

deployment/roles/hub-backend/defaults/main.yml
larsblumberg: What would it take to use `en_US` here? That would be the common setting I guess. Also given that GB left EU...

deployment/roles/hub-develop/tasks/apt,yml
```diff
@@ -0,0 +1,25 @@
...
+
+- name: install rsync
+  apt:
+    name: "{{item}}"
+    state: installed
+  become: true
+  when: source_transport == 'rsync'
+  with_items:
+    - rsync
+
```
larsblumberg: This file uses a comma in its file name.
And @grunskis says, this file shouldn't be even there.

deployment/roles/hub-hass/tasks/main.yml
```diff
@@ -64,12 +64,12 @@
...
-    name: supervisor
-    state: running
-    enabled: true
-  tags: update_hass
+# - name: ensure supervisord is running
+#   service:
+#     name: supervisor
+#     state: running
+#     enabled: true
+#   tags: update_hass
```
larsblumberg: Please add a `TODO` for dead code or delete it

deployment/roles/hub-ubuntu/tasks/main.yml
```diff
@@ -12,41 +12,41 @@
...
+#     - avahi-daemon
+#     - python3
+#     - wpasupplicant
+#     - wireless-tools
+#     - supervisor
+#     - python3-venv
+#     - python3-dbus
+#     - libffi-dev
+#     - libssl-dev
+#     - lighttpd
```
larsblumberg: Please add a `TODO` for dead code or delete it.

Same holds for all consequent dead code in this PR

deployment/roles/hub-ubuntu/tasks/main.yml
```diff
@@ -1,7 +1,7 @@
 ---
-- name: set hostname
-  hostname:
-    name: "{{hostname}}"
+# - name: set hostname
+#   hostname:
+#     name: "{{hostname}}"
```
larsblumberg: Please add a `TODO` for dead code or delete it

deployment/roles/hub-develop/tasks/apt,yml
tomster: i fixed the typo. i vote for leaving the file here, though, because it would make it easier for either us later or users now to configure our stack on their own hardware / platform

that's the reason why i moved this commented out code into a separate playbook in the first place.

deployment/roles/hub-hass/tasks/main.yml
tomster: i'm not sure why this code has been commented out. it wasn't by me, it was @grunskis so we should ask him :)

tomster:
This PR is now good to merge IMHO. i'd be curious to see whether either one of you can bootstrap their instance using this

deployment/roles/hub-hass/tasks/main.yml
grunskis: Just remove it :)

larsblumberg:
I'd like to see the dead code removed before merging

deployment/roles/hub-backend/defaults/main.yml
tomster: I switched the locale both on the OS and in the playbooks

PR #154, opened 2017-05-31
Use wix’s react-native-navigation

larsblumberg:
I kept the original application files for Android and iOS in order we want to revert one day


grunskis:
Just now I realized that I was running an older version of this commit and missed a few things. I committed those changes f9554c1

PR #158, opened 2017-06-02
Add deamon for Wifi onboarding via Bluetooth LE ('bluenetd')

luminosuslight:
To test this on Senic OS run following commands as root:
```
vim /etc/network/interfaces  # comment out wlan0 related lines
killall wpa_supplicant
NetworkManager
cd /srv/senic_hub
venv/bin/bluenetd -b hci0 start -d senic.local
```

Todos in Senic OS:

- running these commands on startup
- fix permission problems for senic user
- remove wlan0 from /etc/network/interfaces
- provide actual DNS name
- remove wifi_scan process (?)

Known issue:

- BLE service is not found on two different devices running Android 5.1 (but Nuimos are working, maybe a problem in how BlueZ advertises the services or a bug in Android)


senic_hub/bluenet/wifi_peripheral.py
```diff
@@ -0,0 +1,287 @@
+#!/usr/bin/python3
```
grunskis: Since these python files are not going to be executable there's no need to have these at the top of the python files...

senic_hub/bluenet/peripheral_bluez.py
```diff
@@ -0,0 +1,243 @@
...
+            self._main_loop.quit()
+            self._main_loop = None
+
+    def add_service(self, service):
+        self._app.add_service(service)
+
+    def add_advertised_service_uuid(self, uuid):
+        self._advertisement.add_service_uuid(uuid)
+
+    # --------------------------------- Private Methods ----------------------------------
```
grunskis: In the rest of the codebase we don't use these kind of markers. I think we should be consistent and remove these...

senic_hub/bluenet/peripheral_bluez.py
```diff
@@ -0,0 +1,243 @@
...
+    def run(self):
+        """
+        Registers advertisement and services to D-Bus and starts the main loop.
+        """
+        if self._main_loop:
+            return
+        self._main_loop = GObject.MainLoop()
+        self._disconnect_all()
+        self._register()
+        print("--- Mainloop started ---")
```
grunskis: Instead of `print` we should use `logging` module or `click.echo`

senic_hub/bluenet/dbus_bluez_interfaces.py
```diff
@@ -0,0 +1,427 @@
...
+                         in_signature='aya{sv}')
+    def WriteValue(self, value, options):
+        self._write_value(value, options)
+
+
+class Advertisement(dbus.service.Object):
+    """
+    org.bluez.LEAdvertisement1 interface implementation
+    """
+    PATH_BASE = '/org/bluez/example/advertisement'
```
grunskis: Is it required to have `example` in this name? If not maybe we could replace it with `bluenet` or something like that

senic_hub/bluenet/dbus_bluez_interfaces.py
```diff
@@ -0,0 +1,427 @@
...
+        return dbus.ObjectPath(self.path)
+
+    def add_characteristic(self, characteristic):
+        self.characteristics.append(characteristic)
+
+    def get_characteristic_paths(self):
+        result = []
+        for chrc in self.characteristics:
+            result.append(chrc.get_path())
+        return result
```
grunskis: I don't mind this verbose syntax, but we're trying to use list/dict/set comprehensions as much as possible. This could be written like `return [c.get_path() for c in self.characteristics]`

senic_hub/bluenet/peripheral_bluez.py
```diff
@@ -0,0 +1,243 @@
...
+
+        dbus.mainloop.glib.DBusGMainLoop(set_as_default=True)
+
+        self.bus = dbus.SystemBus()
+
+        if not adapter:
+            adapter = self._find_adapter()
+            if not adapter:
+                print("Could not find any adapter implementing GattManager1 + LEAdvertisingManager1 interfaces")
+                raise Exception
```
grunskis: You should instantiate Exception when raising: `raise Exception("some message here")`. In fact it's not a good style to raise `Exception` but instead you should define your own `Exception` sub-class and raise that instead...

senic_hub/bluenet/peripheral_bluez.py
```diff
@@ -0,0 +1,243 @@
...
+        if self._main_loop:
+            return
+        self._main_loop = GObject.MainLoop()
+        self._disconnect_all()
+        self._register()
+        print("--- Mainloop started ---")
+        try:
+            self._main_loop.run()
+        except KeyboardInterrupt:
+            pass
```
grunskis: This is raised when you press Ctrl+C? If so then we should probably stop the main loop and exit in this case...

luminosuslight: Yes, it already does this in the finally clause. The empty exception handling just prevents the error message to be written to stdout as we are printing a more meaningful message in the finally clause.

senic_hub/bluenet/peripheral_bluez.py
luminosuslight: IMHO this improves readability greatly and PEP8 kind of suggests this type of markers ("Accordingly, it is important that users be able to clearly distinguish between public and internal interfaces.", and at least for me underscores are not "clearly" visible if I look at the whole class at once). So it would not mean a change of the current code style but a good addition to it in my opinion. Of course I'm open to arguments and can remove it anyway, if requested ;-)

senic_hub/bluenet/peripheral_bluez.py
grunskis: 👍 maybe there could be comment that explains what's happening there

senic_hub/bluenet/onboarding.py
```diff
@@ -177,10 +187,14 @@ def ble_onboarding(ctx, wlan, bluetooth):
...
 @click.option('--alias', '-a', required=False, help="Bluetooth Alias Name (default = HubOnboarding)")
+@click.option('--verbose', '-v', is_flag=True, help="Print verbose messages (log level DEBUG)")
 @click.pass_context
-def ble_onboarding_start(ctx, dns, alias):
+def ble_onboarding_start(ctx, dns, alias, verbose):
     if not alias:
         alias = 'HubOnboarding'
+    if verbose:
+        logging.basicConfig(level=logging.DEBUG)
+    logging.basicConfig(level=logging.DEBUG)
```
grunskis: I guess you wanted to set it to a different level than `DEBUG` in case `-v` wasn't set...

senic_hub/bluenet/peripheral_bluez.py
luminosuslight: Right, added a comment to explain it

senic_hub/bluenet/peripheral_bluez.py
luminosuslight: Maybe you are right about consistency, I removed the separation lines for now

senic_hub/bluenet/onboarding.py
luminosuslight: Yes of course, fixed it

senic_hub/bluenet/peripheral_bluez.py
tomster: FTR i've never really encountered such comments and would also prefer to not have them. having them grouped at the end of the file and prefix them with `_` (as you already do) is sufficient IMHO

senic_hub/bluenet/wifi_peripheral.py
tomster: agreed, this is usually just cargo-cult and often does more harm than good.

senic_hub/bluenet/dbus_bluez_interfaces.py
```diff
@@ -385,7 +385,7 @@ def get_properties(self):
                                                         signature='sv')
         if self.include_tx_power is not None:
             properties['IncludeTxPower'] = dbus.Boolean(self.include_tx_power)
-            return {LE_ADVERTISEMENT_IFACE: properties}
+        return {LE_ADVERTISEMENT_IFACE: properties}
```
tomster: this should really be two separate commits. it's good practice to include whitespace related fixes in a single "cosmetic" commit and separate from the non-cosmetic commits (although in this case the whitespace fix is non-cosmetic, of course)

senic_hub/bluenet/onboarding.py
```diff
@@ -1,13 +1,16 @@
...
-from threading import Thread
 from subprocess import call, CalledProcessError, check_output, STDOUT, TimeoutExpired
+from threading import Thread
 
 from .peripheral_bluez import Peripheral
 from .wifi_peripheral import OnboardingUuids, WifiConnectionState, WifiOnboardingService
 
+logger = logging.getLogger(__name__)
+
 # --------------------------------------------------------------------------
```
tomster: this is superfluous and should be removed.

senic_hub/bluenet/onboarding.py
```diff
@@ -61,8 +64,15 @@ def get_wifi_status(self):
             connection_status = check_output(
                 ['nmcli', '-f', '802-11-wireless.ssid', 'connection', 'show', 'BleOnboarding'],
                 timeout=5, stderr=STDOUT).decode()
-        except (CalledProcessError, TimeoutExpired) as e:
-            print("Wifi status error: %s" % e)
+        except CalledProcessError as e:
+            if e.returncode == 10:
+                # means connection 'BleOnboarding' doesn't exist
```
tomster: I'm glad i caught this, as i renamed it to `senic-office` in the senic-os repo, but of course, in this context it makes more sense to name it like you did, so i will change it back there.

senic_hub/bluenet/wifi_peripheral.py
```diff
@@ -216,7 +217,7 @@ def __init__(self, bus, index, service, dns_name):
     # ------------- Overridden Methods -----------------
```
tomster: hm, i can see how this comment would actually be useful...
i only see the danger of it eventually becoming inaccurate as additional methods are added.
i think, we should leave it here and (at least initially) "tolerate" it as an idiosyncrasy of Tim and perhaps we will get used to it and expand it to other areas :)

senic_hub/bluenet/onboarding.py
```diff
@@ -0,0 +1,208 @@
...
+
+    def join_network(self, ssid, credentials):
+        logger.info("Trying to join network: %s, Credentials: %s" % (ssid, credentials))
+        self._configure_wlan(ssid, credentials)
+
+    def get_wifi_status(self):
+        # get current SSID:
+        try:
+            connection_status = check_output(
+                ['nmcli', '-f', '802-11-wireless.ssid', 'connection', 'show', 'BleOnboarding'],
```
tomster: @luminosuslight does this mean, that the OS must always provide a connection with this name, even if it initially isn't configured?
currently i only add this connection for the `-dev` version of senicOS but if i understand this correctly, we should do this also for production, right? please confirm...

luminosuslight: Ah no, the connection "BleOnboarding" doesn't have to exist on startup, the code can handle this and creates it when necessary.
You're totally right that this is not obvious and I just made a constant for this connection name and added an explaining comment to it.

senic_hub/bluenet/onboarding.py
luminosuslight: I made a constant for the connection name to be able to change it and make sure that it is the same in all places.

senic_hub/bluenet/onboarding.py
luminosuslight: Do you mean the 'logger' line or the comment below? (I removed the comment line in b4effa2)

senic_hub/bluenet/wifi_peripheral.py
luminosuslight: I removed the comment lines in b4effa2.

senic_hub/bluenet/dbus_bluez_interfaces.py
luminosuslight: Yeah I know, don't know why I mixed this up here (the other commits could be more fine grained, too, I just realize)

deployment/fab_pi3.py
```diff
@@ -119,36 +119,3 @@ def sync_src():
...
-        rsync(
-            '-rlptvD',
-            '--exclude', '.*',
-            '--exclude', '*.egg-info',
-            '--exclude', '__pycache__',
-            '--exclude', '/dist',
-            '--exclude', '/docs',
-            '--exclude', '/venv',
-            '.',
-            '{host_string}:%s' % destination)
```
larsblumberg: Why are these synching methods being removed? I understand that we move to Nano Pi as a development platform but then we should remove the whole `fab_pi3.py` file. If we continue to develop on Pi3 (I still do), then these sync methods are helpful.

senic_hub/bluenet/dbus_bluez_interfaces.py
```diff
@@ -0,0 +1,395 @@
...
+    _dbus_error_name = 'org.bluez.Error.InvalidValueLength'
+
+
+class FailedException(dbus.exceptions.DBusException):
+    _dbus_error_name = 'org.bluez.Error.Failed'
+
+
+def string_to_dbus_array(value):
+    arr = [dbus.Byte(c) for c in value.encode()]
+    return arr
```
larsblumberg: `return [dbus.Byte(c) for c in value.encode()]`

senic_hub/bluenet/dbus_bluez_interfaces.py
```diff
@@ -0,0 +1,395 @@
...
+
+class InvalidValueLengthException(dbus.exceptions.DBusException):
+    _dbus_error_name = 'org.bluez.Error.InvalidValueLength'
+
+
+class FailedException(dbus.exceptions.DBusException):
+    _dbus_error_name = 'org.bluez.Error.Failed'
+
+
+def string_to_dbus_array(value):
```
larsblumberg: Since its converting a string array, I suggest to name it `string_array_to_dbus_array` because `to` methods tend to be specific about what they convert.

senic_hub/bluenet/dbus_bluez_interfaces.py
```diff
@@ -0,0 +1,395 @@
...
+    arr = [dbus.Byte(c) for c in value.encode()]
+    return arr
+
+
+class Application(dbus.service.Object):
```
larsblumberg: We agreed on ordering classes and methods by their usage from the outside. Doing so helps the reader to get a bird perspective first before going into detail.

In this file it's mostly the case, class definitions start with `Application`, followed by `Service`, `Characteristic`, etc. All exceptions and functions listed above should go, according to our convention, to the very end of this file as they are the smallest details when reading and understanding this file.

senic_hub/bluenet/dbus_bluez_interfaces.py
```diff
@@ -0,0 +1,395 @@
...
+        super().__init__(bus, self.path)
+
+    def get_path(self):
+        return dbus.ObjectPath(self.path)
+
+    def add_service(self, service):
+        self.services.append(service)
+
+    @dbus.service.method(DBUS_OM_IFACE,
+                         out_signature='a{oa{sa{sv}}}')
```
larsblumberg: When we indent, then we use only 4 spaces, we don't align to opening brackets. Here I would even prefer to read it all one line since it's short enough.

senic_hub/bluenet/dbus_bluez_interfaces.py
```diff
@@ -0,0 +1,395 @@
...
+    def GetManagedObjects(self):
+        response = {}
+
+        for service in self.services:
+            response[service.get_path()] = service.get_properties()
+
+            chrcs = service.get_characteristics()
```
larsblumberg: I personally I am not a fan of unknown abbreviations such as `chrcs`. I'm okay with that but I would prefer to read what it is, an array of `characteristics`.

senic_hub/bluenet/dbus_bluez_interfaces.py
```diff
@@ -0,0 +1,395 @@
...
+            for chrc in chrcs:
```
larsblumberg: Same for `chrc`, I suggest to name it `characteristic` or, also fine, `c`.

For the sake of pragmatism you can leave it as it is and simply pay attention for the next time similar code needs to be written.

senic_hub/bluenet/dbus_bluez_interfaces.py
```diff
@@ -0,0 +1,395 @@
...
+class Service(dbus.service.Object):
+    """
+    org.bluez.GattService1 interface implementation
+    
+    This is intended to be subclassed with a concrete implementation.
+    """
+    PATH_BASE = '/org/bluez/example/service'
+
+    def __init__(self, bus, index, uuid, primary):
+        self.path = self.PATH_BASE + str(index)
```
larsblumberg: I would inline the value of `PATH_BASE` since it's only used once. Having this constant adds unnecessary cognitive load.

senic_hub/bluenet/dbus_bluez_interfaces.py
```diff
@@ -0,0 +1,395 @@
...
+        self.bus = bus
+        self.uuid = uuid  # string
+        self.primary = primary  # bool
+        self.characteristics = []
+        super().__init__(bus, self.path)
```
larsblumberg: Why is `super().init` called at the end? This atypical – Does the implementation of `dbus.service.Object` require it?

senic_hub/bluenet/dbus_bluez_interfaces.py
```diff
@@ -0,0 +1,395 @@
...
+        }
+
+    def get_path(self):
+        return dbus.ObjectPath(self.path)
+
+    def add_characteristic(self, characteristic):
+        self.characteristics.append(characteristic)
+
+    def get_characteristic_paths(self):
+        return [chrc.get_path() for chrc in self.characteristics]
```
larsblumberg: We agreed on using single letter variables in comprehensions, namely the first letter of the variable name, if it would be fully spelled. Here: `c`. Sorry for not having yet written down our conventions, right now they live only from mouth to mouth.

senic_hub/bluenet/dbus_bluez_interfaces.py
```diff
@@ -0,0 +1,395 @@
...
+        """
+        This is a newly introduced method that should be called when the remote device disconnected.
+        This is necessary to be able to stop notifying (BlueZ doesn't call StopNotify when device disconnects).
+        """
+        for chrc in self.characteristics:
+            chrc.remote_disconnected()
+
+    @dbus.service.method(DBUS_PROP_IFACE,
+                         in_signature='s',
+                         out_signature='a{sv}')
```
larsblumberg: According to our conventions we use a one-liner here or alternatively:

    @dbus.service.method(
        DBUS_PROP_IFACE,
        in_signature='s',
        out_signature='a{sv}')

senic_hub/bluenet/dbus_bluez_interfaces.py
```diff
@@ -0,0 +1,395 @@
...
+    def GetAll(self, interface):
+        if interface != GATT_SERVICE_IFACE:
+            raise InvalidArgsException()
+
+        return self.get_properties()[GATT_SERVICE_IFACE]
+
+    @dbus.service.signal(DBUS_PROP_IFACE,
+                         signature='sa{sv}as')
```
larsblumberg: Indentation

senic_hub/bluenet/dbus_bluez_interfaces.py
```diff
@@ -0,0 +1,395 @@
...
+        self.service_data = None
+        self.include_tx_power = False
+        super().__init__(bus, self.path)
+
+    def get_properties(self):
+        properties = dict()
+        properties['Type'] = self.ad_type
+        if self.service_uuids is not None:
+            properties['ServiceUUIDs'] = dbus.Array(self.service_uuids,
+                                                    signature='s')
```
larsblumberg: Indentation doesn't correspond to our conventions. Please also check following code.

senic_hub/bluenet/onboarding.py
```diff
@@ -0,0 +1,213 @@
...
+
+#: Name of the NetworkManager connection to use for BLE onboarding
+#: It will be created on first attempt to join a network
+#: or overwritten if it already exists.
+NM_CONNECTION_NAME = 'BleOnboarding'
+
+logger = logging.getLogger(__name__)
+
+
+class OnboardingDaemon(object):
```
larsblumberg: What about naming this class `BluenetDaemon` and the file `bluenet.py`? Since we want to open-source this module I expected `onboarding` not be understood by every reader.

senic_hub/bluenet/onboarding.py
```diff
@@ -0,0 +1,213 @@
...
+from subprocess import call, CalledProcessError, check_output, STDOUT, TimeoutExpired
+from threading import Thread
+
+from .peripheral_bluez import Peripheral
+from .wifi_peripheral import OnboardingUuids, WifiConnectionState, WifiOnboardingService
+
+#: Name of the NetworkManager connection to use for BLE onboarding
+#: It will be created on first attempt to join a network
+#: or overwritten if it already exists.
+NM_CONNECTION_NAME = 'BleOnboarding'
```
larsblumberg: Better: `NM_CONNECTION_NAME = 'Bluenet'`?

senic_hub/bluenet/onboarding.py
```diff
@@ -0,0 +1,213 @@
...
+            try:
+                iw_scan_result = check_output(['iw', 'dev', self._wlan_adapter, 'scan']).decode()
+            except CalledProcessError as e:
+                logger.warning("Scanning wifi networks failed: %s" % e)
+                return
+
+            networks = []
+            for line in iw_scan_result.split('\n'):
+                if "SSID: " in line:
+                    ssid = line.split(': ')[1]
```
larsblumberg: If SSID name contains the string `: `, using split() as such will not extract the full SSID name. Better split only at first occurrence and including `SSID: ` text.

Better: `ssid = line.split('SSID: ', 1)[1]`

Alternatively use regex, then `if in` becomes obsolete.

senic_hub/bluenet/onboarding.py
```diff
@@ -0,0 +1,213 @@
...
+        def scan_wifi_networks():
+            try:
+                iw_scan_result = check_output(['iw', 'dev', self._wlan_adapter, 'scan']).decode()
+            except CalledProcessError as e:
+                logger.warning("Scanning wifi networks failed: %s" % e)
+                return
+
+            networks = []
+            for line in iw_scan_result.split('\n'):
+                if "SSID: " in line:
```
larsblumberg: We use double quotes to encode text that is printed to humans and for string that can be simply changed without any worries because it isn't referenced anywhere. Here we are looking for precisely the string `SSID ` (it acts as an identifier) so we encode it using single quotes.

senic_hub/bluenet/onboarding.py
```diff
@@ -0,0 +1,213 @@
...
+                if ssid not in found_networks:
+                    logger.info("New network found: %s" % ssid)
+                found_networks[ssid] = time.time()
+
+            # Remove networks that we didn't see for a while
+            now = time.time()
+            for ssid, last_seen in found_networks.copy().items():
+                if now - last_seen > discard_time:
+                    logger.info("Network disappeared: %s" % ssid)
+                    del found_networks[ssid]
```
larsblumberg: I prefer when same things are named same, here I would prefer to read `delete found_ssids[ssid]` or `delete found_networks[network]`. It's okay to leave it now, we should however pay more attention to those details as using different names increases cognitive load.

senic_hub/bluenet/onboarding.py
```diff
@@ -0,0 +1,213 @@
...
+            call(['nmcli', 'dev', 'wifi', 'con', ssid, 'ifname', self._wlan_adapter, 'name', NM_CONNECTION_NAME])
+
+        self._is_joining_wifi = False
+
+
+@click.group()
+@click.pass_context
+@click.option('--wlan', '-w', required=False, help="WLAN device (default = wlan0)")
+@click.option('--bluetooth', '-b', required=False, help="Bluetooth device")
+def ble_onboarding(ctx, wlan, bluetooth):
```
larsblumberg: This method is the main entry point of this file. This said, all methods belonging to the command line tools should come first. Then after these methods the class definition of `OnboardingDaemon` (or `BluenetDaemon`) follows. Again, we agreed on such an order because each file tells a story. This story here starts with the main entry point `def ble_onboarding`.

senic_hub/bluenet/onboarding.py
```diff
@@ -0,0 +1,213 @@
...
+
+
+@ble_onboarding.command(name='start', help="start GATT service and scan for networks")
+@click.option('--dns', '-d', required=True, help="DNS Name of Hub")
+@click.option('--alias', '-a', required=False, help="Bluetooth Alias Name (default = HubOnboarding)")
+@click.option('--verbose', '-v', is_flag=True, help="Print verbose messages (log level DEBUG)")
+@click.pass_context
+def ble_onboarding_start(ctx, dns, alias, verbose):
+    if not alias:
+        alias = 'HubOnboarding'
```
larsblumberg: Either make `alias` a required parameter or don't require it all. It leads to bugs if we fall back to this name in case somebody doesn't specify it. This said we now have 2 points of truth where we define the advertisement name. One is here and at least another is there where we call this tool.

senic_hub/bluenet/peripheral_bluez.py
```diff
@@ -0,0 +1,251 @@
...
+
+        self.alias = alias
+        self.is_powered = True
+        self.is_discoverable = True
+        self.discoverable_timeout = 0
+
+        # Prepare Managers:
+
+        self._ad_manager = dbus.Interface(self.bus.get_object(BLUEZ_SERVICE_NAME, self._adapter_path),
+                                          LE_ADVERTISING_MANAGER_IFACE)
```
larsblumberg: Indentation not corresponding to our convention and following

senic_hub/bluenet/peripheral_bluez.py
```diff
@@ -0,0 +1,251 @@
...
+        except KeyboardInterrupt:
+            # ignore exception as it is a valid way to exit the program
+            # and skip to finally clause
+            pass
+        except Exception as e:
+            logger.error(e)
+        finally:
+            logger.info("--- Mainloop finished ---")
+            self._unregister()
+            self._main_loop.quit()
```
larsblumberg: And `self._main_loop = None` as in `def stop`?

senic_hub/bluenet/peripheral_bluez.py
```diff
@@ -0,0 +1,251 @@
...
+
+    def stop(self):
+        """
+        Stops the main loop started with `run()`
+        """
+        if self._main_loop:
+            self._unregister()
+            self._main_loop.quit()
+            self._main_loop = None
```
larsblumberg: Here's code duplication with `def run`. Wouldn't it be sufficient if only `self._main_loop.quit()` would be invoked? After that the `finally` block of `def start` would be run, doing exactly the same.

senic_hub/bluenet/peripheral_bluez.py
```diff
@@ -0,0 +1,251 @@
...
+    def add_advertised_service_uuid(self, uuid):
+        self._advertisement.add_service_uuid(uuid)
+
+    def _find_adapter(self):
+        """
+        Tries to find a bluetooth adapter with the required functions.
+        :return: path to an adapter or None
+        """
+        required_interfaces = [GATT_MANAGER_IFACE, LE_ADVERTISING_MANAGER_IFACE]
+        remote_om = dbus.Interface(self.bus.get_object(BLUEZ_SERVICE_NAME, '/'), DBUS_OM_IFACE)
```
larsblumberg: I find `object_manager` a simpler name to understand instead of `remote_om`.

senic_hub/bluenet/peripheral_bluez.py
```diff
@@ -0,0 +1,251 @@
...
+        objects = remote_om.GetManagedObjects()
+
+        for object_path, properties in objects.items():
+            interface_missing = False
+            for interface in required_interfaces:
+                if interface not in properties.keys():
+                    interface_missing = True
```
larsblumberg: Use list comprehensions where applicable:

    missing_interface = next([i for i in required_interfaces if interface not in properties.keys()], None)
    if missing_interface:
        continue

Please check the code if it makes sense, I didn't test it.

senic_hub/bluenet/wifi_peripheral.py
```diff
@@ -0,0 +1,266 @@
...
+
+
+class WifiConnectionState(Enum):
+    DOWN = 0
+    DISCONNECTED = 1
+    CONNECTING = 2
+    CONNECTED = 3
+
+
+class WifiOnboardingService(Service):
```
larsblumberg: Let's brand this module everywhere and name this `class BluenetService`

senic_hub/bluenet/wifi_peripheral.py
```diff
@@ -0,0 +1,266 @@
...
+        It must have the signature callback(ssid, credentials).
+        """
+        self._join_callback = cb
+
+    def _credentials_received(self):
+        if callable(self._join_callback):
+            self._join_callback(self._ssid_chrc.ssid, self._credentials_chrc.credentials)
+        
+        
+class AvailableNetworksChrc(Characteristic):
```
larsblumberg: No abbreviations please: `AvailableNetworksCharacteristic`, also please rename other classes

senic_hub/bluenet/wifi_peripheral.py
```diff
@@ -0,0 +1,266 @@
...
+    def _start_notify(self):
+        logger.info("Enabled notification about connection state.")
+        self._notifying = True
+
+    def _stop_notify(self):
+        logger.info("Disabled notification about connection state.")
+        self._notifying = False
+
+
+class DnsNameChrc(Characteristic):
```
larsblumberg: If you have time, please rename every usage of "DNS name" to `host`. It's also likely that we notify an IP address instead of a DNS name. Using `host` allows for larger flexibility. I know that the DNS name could theoretically be also an IP address, but renaming it to "host" makes it obvious. That means, that the `-d` parameter for bluenet becomes a `-h` parameter.

senic_hub/bluenet/wifi_peripheral.py
```diff
@@ -0,0 +1,266 @@
...
+    def __init__(self, bus, index, service):
+        super().__init__(bus, index, OnboardingUuids.CREDENTIALS, ['write'], service)
+        self.credentials = None
+        self.callback = None
+
+    def set_callback_on_change(self, cb):
+        self.callback = cb
+
+    def _write_value(self, value, options):
+        self.credentials = ''.join([chr(byte) for byte in value])
```
larsblumberg: Instead of `''.join(...)`, would `str([chr(byte) for byte in value])` also work? If yes, it seems more straight forward to me.

senic_hub/bluenet/dbus_bluez_interfaces.py
luminosuslight: 'string_array' sounds to me like an array of strings while this method converts a single Python3 unicode string to a tuple of dbus.Bytes. Should I change it anyway?

senic_hub/bluenet/peripheral_bluez.py
luminosuslight: I replaced it with
`missing_interfaces = [i for i in required_interfaces if i not in properties.keys()]
if missing_interfaces:
    continue`

senic_hub/bluenet/wifi_peripheral.py
luminosuslight: `str(tuple)` returns a human readable representation (i.e. `"['A', 'B', 'C']"`). I replaced it with `bytes(value).decode()` instead.

deployment/fab_pi3.py
luminosuslight: Reverted this in 7635f18

senic_hub/bluenet/dbus_bluez_interfaces.py
luminosuslight: Fixed in 32cb500

senic_hub/bluenet/dbus_bluez_interfaces.py
luminosuslight: Reordered files in dc1de26

senic_hub/bluenet/dbus_bluez_interfaces.py
luminosuslight: Removed abbreviations in 0900af2

senic_hub/bluenet/dbus_bluez_interfaces.py
luminosuslight: The constant exists to be able to override it in subclasses. I added a comment to make this clear in 74835fa.

luminosuslight:
Addressed all requested changes (indentation, renaming, code style). Renamed files and branded everything 'bluenet', too.

senic_hub/bluenet/bluenet.py
```diff
@@ -0,0 +1,212 @@
...
+
+@bluenet_cli.command(name='status', help="print Wifi status")
+@click.pass_context
+def bluenet_status(ctx):
+    print(ctx.obj.get_wifi_status())
+
+
+class BluenetDaemon(object):
+    """
+    Daemon to enable Wifi onboarding via Bluetooth Low Energy.
```
larsblumberg: The correct terminus would be Wi-Fi provisioning – no need to change now

senic_hub/bluenet/dbus_bluez_interfaces.py
larsblumberg: My bad!

PR #161, opened 2017-06-07
Boot screen

grunskis:
* Reads hub host from the device and store in on device
* Implement screen that's launched first that checks if the hub has been on-boarded and is reachable

### DIFF
--- a/senic_hub/setup_app/components/BootScreen.js
+++ b/senic_hub/setup_app/components/BootScreen.js
@@ -1,0 +1,98 @@
+import React from 'react';
+import {
+  ActivityIndicator,
+  Button,
+  StyleSheet,
+  Text,
+  View,
+} from 'react-native';
+import Screen from './Screen.js';
+import Settings from '../Settings'
+
+export default class BootScreen extends Screen {
+  constructor(props) {
+    super(props)
+
+    this.state = {
+      hubUnreachable: false,
+    }
+  }
+
+  didAppear() {
+    this.pingHub()
+  }
+
+  pingHub() {
+    Settings.getHubApiUrl()
+      .then(hubApiUrl => {
+        if (hubApiUrl) {
+          console.log('Hub', hubApiUrl, 'already onboarded...')
+
+          fetch(hubApiUrl)
+            .then(() => {
+              Settings.HUB_API_URL = hubApiUrl  // TODO is there a better way?
+
+              console.log('Host', Settings.HUB_API_URL, 'reachable...')
+
+              this.resetTo('app.nuimoComponents')
+            })
+            .catch(() => this.setState({hubUnreachable: true}))
+        }
+        else {
+          this.resetTo('setup.welcome')
+        }
+      })
+      .catch(error => {
+        console.warn(error)
+
+        this.resetTo('setup.welcome')
+      })
+  }
+
+  render() {
+    return (
+      <View style={styles.container}>
+        {this._renderContent()}
+      </View>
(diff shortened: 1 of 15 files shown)


senic_hub/setup_app/components/BootScreen.js
```diff
@@ -0,0 +1,98 @@
...
+      this.setTitle('Oh, noes')
+
+      return (
+        <View>
+          <View>
+            <Text>
+              We couldn't reach your Senic hub. Make sure it's connected to power and within reach of your Wi-Fi network.
+            </Text>
+            <Text>
+              In case the onboarding process fails, please try powering on/off the device. If that doesn't help, then perform the factory reset of the device.
```
larsblumberg: Nice hint to the user!

senic_hub/setup_app/components/BootScreen.js
```diff
@@ -0,0 +1,98 @@
...
+    this.pingHub()
+  }
+
+  pingHub() {
+    Settings.getHubHost()
+      .then(hubHost => {
+        if (hubHost) {
+          console.log('Hub', hubHost, 'already onboarded...')
+
+          fetch(hubHost + '/-/')
```
larsblumberg: Does `Settings.getHubHost()` also return the protocol? If that's the case IMHO `Settings.getHubHost()` should only return the hub host, i.e. `pi3martins.local` without the protocol, not `http://pi3martins.local`.

senic_hub/setup_app/components/DeviceSelection.js
```diff
@@ -65,7 +65,7 @@ export default class DeviceSelection extends Screen {
 
   fetchComponent() {
     let that = this
-    return fetch(API_URL + '/-/nuimos/0/components/' + that.state.component.id)
+    return fetch(Settings.HUB_HOST + '/-/nuimos/0/components/' + that.state.component.id)
```
larsblumberg: In case `Settings.HUB_HOST` also contains the protocol, we want to rename `Settings.HUB_HOST` to `Settings.HUB_URL`

senic_hub/setup_app/components/Screen.js
```diff
@@ -81,6 +81,10 @@ export default class Screen extends Component {
     }
   }
 
+  reset(screen) {
```
larsblumberg: I would name this method to `resetTo` as it doesn't reset the given screen put it replaces the whole navigation stack with a new root screen. Alternatively the method could also be named `setRootScreen()`

senic_hub/setup_app/components/BootScreen.js
grunskis: Yes, returns protocol & port. Client then can take the string and not worry about anything (for example if we decide one day to switch to `https` or change the port number we wouldn't need to release new client versions). What do you think about this approach?

senic_hub/setup_app/components/DeviceSelection.js
grunskis: Maybe even `HUB_API_URL`?

senic_hub/setup_app/components/Screen.js
grunskis: +1 for `resetTo` to match method for wix navigation.

senic_hub/setup_app/components/DeviceSelection.js
larsblumberg: Good idea!

PR #169, opened 2017-06-12
Make onboarding more robust

larsblumberg:
@grunskis The 3 things we paired on:
- Don't send `disconnected` state while hub is connecting to a wifi network
- Report `onboarded` in app info endpoint and use it to decide by app if hub is onboarded
- Move "join wifi" button into navigation bar


senic_hub/bluenet/bluenet.py
```diff
@@ -182,12 +182,24 @@ def print_status(status):
...
                 print_status(new_status)
+                if (self._is_joining_wifi and
+                        new_status == WifiConnectionState.DISCONNECTED and
+                        time.time() - last_status_changed_time < 5.0):
+                    logger.info("Ignoring state change to DISCONNECTED as we are trying to connect to a network")
+                    # When WiFi adapter was already connected, the WiFi state will change
+                    # from `connecting` to `disconnected` to `connected`. To prevent
+                    # `disconnected` to be notified, we wait 5 seconds if we actually
+                    # succeed to connect after a intermediate disconnect step.
+                    continue
```
luminosuslight: Wouldn't this `continue` lead to time.sleep(1) not being executed? A solution could be to put the next three lines in an `else` statement.

larsblumberg: Very good catch! Will use an `else` case instead of `continue`

larsblumberg: @luminosuslight Addressed in e940f79

Issue #130, opened 2017-04-27
Supervisor program logs fill up disk space


grunskis:
Particularly Nuimo app & Device discovery program logs pile up quite quickly.

Limit the number of rotated files that are kept around.

tomster:
implemented in https://github.com/getsenic/senic-hub/pull/176

PR #176, opened 2017-06-14
#130 log rotation

tomster:
unifies the log handling and limits it to 50mb to avoid filling up the card


deployment/roles/hub-backend/templates/supervisor.conf
```diff
@@ -9,8 +9,12 @@ directory={{ backend_deploy_location }}
 {% endif %}
 autostart=true
 autorestart=true
-stdout_logfile={{ backend_data_location }}/wsgi.log
-redirect_stderr=true
+stdout_logfile={{ backend_data_location }}/senic_hub.log
```
larsblumberg: +1 for renaming the log file to `senic_hub.log`

grunskis:
@tomster Until now all logs were written to a single file. I noticed only now that you introduced another file for `error` logging. Why there's need for that?

tomster:
@grunskis it's a technical reason. since we want to use syslog, we will need to hook into supervisor's event machinery and the docs say "Setting redirect_stderr=true in an [eventlistener:x] section is now disallowed because any messages written to stderr would interfere with the eventlistener protocol on stdout"

tomster:
but now that i re-read that, i think it doesn't actually apply in this case, since we're not setting the logging in the listener, but rather in the application.

so there's no reason not to keep it

PR #177, opened 2017-06-14
Deploy iOS test builds to Crashlytics

grunskis:
Followed this tutorial for the most part https://medium.com/react-native-training/fastlane-for-react-native-ios-android-app-devops-8ca85bee614e


senic_hub/setup_app/ios/SenicHubSetup/Info.plist
```diff
@@ -51,6 +51,8 @@
 			</dict>
 		</dict>
 	</dict>
+	<key>NSLocationWhenInUseUsageDescription</key>
```
larsblumberg: We don't want to track the user's location. We shouldn't have this key here. Also next line `<string></string>` needs to be removed as it belongs together.

PR #179, opened 2017-06-15
Make wifi onboarding more robust

larsblumberg:


senic_hub/bluenet/dbus_bluez_interfaces.py
```diff
@@ -189,10 +189,12 @@ def _stop_notify(self):
...
         """
-        This is a newly introduced method that should be called when the remote device disconnected.
-        This is necessary to be able to stop notifying (BlueZ doesn't call StopNotify when device disconnects).
-        It doesn't call _stop_notify() directly because it can not know if this characteristic supports notifying at all.
+        Called when the remote device disconnected.
+        Should be overridden to stop notifying characteristic changes,
+        because BlueZ doesn't call StopNotify when the remote device disconnects.
         """
+
+        # TODO: We can call _stop_notify() for all characteristics here.
```
luminosuslight: At the moment this would lead to a NotSupportedException if the characteristic doesn't support notifications. A solution could be to remove the warning and exception in the default _stop_notify() method.

senic_hub/setup_app/components/SetupWifiPassword.js
```diff
@@ -51,33 +54,77 @@ export default class SetupWifiPassword extends Screen {
...
+      })
+      .catch((error) => {
+        this.setState({isSendingPassword: false})
+        console.error(error)
+      })
+  }
+
+  continue() {
+    // TODO: We have to read hub's host name from the hub only after it connected
+    //       to wi-fi network, because the hostname depends on which IP address it got
```
luminosuslight: I will enable notification support for the host name characteristic in the next pull request, so that the host name value is always up-to-date (if it is subscribed to).

PR #181, opened 2017-06-16
Improve Bluenet (auto advertise, %IP placeholder, etc.)

luminosuslight:
- disable advertisement when not needed and reactivate it when Wifi connection is lost if --auto-advertise flag is provided
- support %IP placeholder in hostname
- only log Wifi password when debugging (with -vv option)
- force reconnecting to Wifi network if it appears again after disappearing
- clean up stop notify mechanism
- join Wifi in new thread to not block bluetooth thread


senic_hub/bluenet/bluenet.py
```diff
@@ -90,14 +103,13 @@ def run(self, hostname, bluetooth_alias):
...
     def join_network(self, ssid, credentials):
-        # TODO: Join Wi-Fi on a separate thread as otherwise the Bluetooth
-        #       thread gets block and no write response is sent for the
-        #       crendentials characteristic.
-        logger.info("Trying to join network: %s, Credentials: %s" % (ssid, credentials))
-        self._configure_wlan(ssid, credentials)
+        logger.info("Trying to join network: %s" % ssid)
+        logger.debug("Password: %s" % credentials)
+        thread = Thread(target=self._configure_wlan, args=(ssid, credentials))
+        thread.start()
```
larsblumberg: Please leave a comment here why we are starting a thread.

Please check if the thread was already started, then don't join again. Otherwise the central could invoke another thread with sending another password while the previous network join still runs.

senic_hub/bluenet/bluenet.py
```diff
@@ -199,10 +222,17 @@ def print_status(status):
...
                     last_status_changed_time = time.time()
-            time.sleep(1)
+            if self._ble_peripheral.is_connected:
+                time.sleep(1)
+            else:
+                # NetworkManager is slow at detecting connection changes
+                # -> checking every 30s is enough while setup app is not connected
+                # (when Wifi router is turned off connection status will be "Connecting"
+                # for 2 minutes before going to "Disconnected")
+                time.sleep(30)
```
larsblumberg: Super!

senic_hub/bluenet/bluenet.py
```diff
@@ -230,6 +260,60 @@ def _configure_wlan(self, ssid, password):
...
+        if (new_status == WifiConnectionState.CONNECTED and
+                not self._ble_peripheral.is_connected and
+                self._ble_peripheral.is_advertising and
+                not self._keep_advertising):
+            # re-enabling advertisement is not possible while a device is connected
+            # because of that we are not disabling it in the first place when a device is connected
+            logging.info("Wifi reconnected. Stopping BLE advertisement.")
+            self._ble_peripheral.stop_advertising()
+        elif new_status == WifiConnectionState.DISCONNECTED \
+                and not self._ble_peripheral.is_advertising:
```
larsblumberg: Please use brackets instead of `\`, this way we avoid too many different coding styles.

senic_hub/bluenet/bluenet_gatt_service.py
```diff
@@ -190,25 +190,37 @@ def _stop_notify(self):
...
-    def remote_disconnected(self):
-        self._stop_notify()
-
 
 class HostNameCharacteristic(Characteristic):
     """
     GATT characteristic providing the host name of the server.
-
-    Possible operations: Read
+
```
larsblumberg: You can configure your editor to not have lines with trailing spaces

senic_hub/bluenet/bluez_peripheral.py
```diff
@@ -234,15 +259,16 @@ def _register_app_error_cb(self, error):
     def _register_ad_error_cb(self, error):
```
larsblumberg: Name of this method doesn't match the naming pattern of other callbacks, i.e. `_adapter_properties_changed`. We could use `_register_advertisement_failed()` here to match our naming pattern.

senic_hub/bluenet/bluenet_gatt_service.py
```diff
@@ -190,25 +190,37 @@ def _stop_notify(self):
...
-        return string_to_dbus_array(self.host_name)
+        return string_to_dbus_array(self.hostname)
+
+    def _start_notify(self):
+        logger.info("Enabled notification about hostname.")
+        self._notifying = True
+
+    def _stop_notify(self):
+        logger.info("Disabled notification about hostname.")
+        self._notifying = False
```
larsblumberg: Can we move these two methods `_start_notify` and `_stop_notify` into the base class, as well as the attribute `self._notifying`? Also, does it make sense to then remove the leading underscore as it wouldn't be really "private" then? Should it be private?

senic_hub/bluenet/bluenet_gatt_service.py
luminosuslight: Check ✔

senic_hub/bluenet/bluenet_gatt_service.py
luminosuslight: Moved it to the base class as default implementations but left the underscore because those methods are still never called from the outside (only inside inheritance hierarchy).

luminosuslight:
Addressed all requested changes.

senic_hub/bluenet/bluenet_gatt_service.py
luminosuslight: Actually I don't know if the underscore in Python corresponds to private or protected in C++, do you know?

larsblumberg: A single underscore roughly corresponds to _protected_ and a double underscore encodes roughly the semantics of _private_ like private in C-like languages, ignoring the actual accessibility of course.

+1 for moving it into the base class. I also now understand that it's only used inside inheritance hierarchy, so I see why you keep the underscore.

senic_hub/bluenet/bluenet.py
```diff
@@ -70,17 +73,27 @@ def __init__(self, wlan_adapter, bluetooth_adapter):
...
-        gatt_service = BluenetService(peripheral.bus, 0, hostname, "1.0")
+        self._ble_peripheral = Peripheral(bluetooth_alias, self._bluetooth_adapter)
+        gatt_service = BluenetService(self._ble_peripheral.bus, 0, self._get_hostname(), "1.0")
         gatt_service.set_credentials_received_callback(self.join_network)
         self._gatt_service = gatt_service
-        peripheral.add_service(gatt_service)
-        peripheral.add_advertised_service_uuid(BluenetUuids.SERVICE)
+        self._ble_peripheral.add_service(gatt_service)
+        self._ble_peripheral.add_advertised_service_uuid(BluenetUuids.SERVICE)
+        self._ble_peripheral.on_remote_disconnected(self._update_advertising_state)
```
larsblumberg: As we just discussed with @grunskis, we want to directly assign the callback to a "public" instance variable of ble peripheral.

senic_hub/bluenet/bluez_peripheral.py
```diff
@@ -107,6 +108,30 @@ def add_service(self, service):
     def add_advertised_service_uuid(self, uuid):
         self._advertisement.add_service_uuid(uuid)
 
+    def on_remote_disconnected(self, cb):
+        self._remote_disconnected_callback = cb
+
+    def start_advertising(self):
```
larsblumberg: This method might be called multiple times from `_update_advertising_state`. We don't check here if we're  already advertising. Is this harmful? I.e. is it harmful to do `self._ad_manager.RegisterAdvertisement` multiple times?

senic_hub/bluenet/bluez_peripheral.py
```diff
@@ -170,7 +195,14 @@ def is_powered(self, value):
...
     def is_connected(self):
-        return self._adapter_props.Get('org.bluez.Device1', 'Connected') == 1
+        object_manager = dbus.Interface(self.bus.get_object(BLUEZ_SERVICE_NAME, '/'), DBUS_OM_IFACE)
+        objects = object_manager.GetManagedObjects()
+
+        for object_path, properties in objects.items():
+            if 'org.bluez.Device1' in properties.keys():
+                if properties['org.bluez.Device1'].get('Connected', False):
+                    return True
+        return False
```
larsblumberg: This can be expressed via a comprehension:

    connected_devices = [
        object_path for object_path, properties in objects.items()
        if properties.get('org.bluez.Device1', {}).get('Connected', False)
    ]
    return len(connected_devices) > 0

senic_hub/bluenet/bluez_peripheral.py
larsblumberg: This method might be called multiple times from `_update_advertising_state`. We don't check here if we're  already advertising. Is this harmful? I.e. is it harmful to do `self._ad_manager.RegisterAdvertisement` multiple times?

senic_hub/bluenet/bluenet.py
larsblumberg: As we just discussed with @grunskis, we prefer to directly assign the callback to a "public" instance variable of ble peripheral. I.e.:

    self._ble_peripheral.on_remote_disconnected = self._update_advertising_state

senic_hub/bluenet/dbus_bluez_interfaces.py
```diff
@@ -172,30 +177,25 @@ def _write_value(self, value, options):
...
         """
         Called when the remote device disconnected.
-        Should be overridden to stop notifying characteristic changes,
+        Calls _stop_notify() to stop notifying characteristic changes,
         because BlueZ doesn't call StopNotify when the remote device disconnects.
         """
-
-        # TODO: We can call _stop_notify() for all characteristics here.
-        pass
+        self._on_stop_notifying()
```
larsblumberg: This line should be `self._stop_notify()`, otherwise we wouldn't reset the `self._is_notifying` state.

larsblumberg:
Great PR! Still we found a few things, could you please look into them?

grunskis:
Nice stuff! Can't wait for this to get merged to get the `%IP` placeholder feature :)

larsblumberg:
@luminosuslight: We're merging your branch now since it generally works. However, let's address the open issues in this PR on Friday when you are here.

PR #194, opened 2017-06-23
Add 'netwatch' supervisor

luminosuslight:
Adds a supervisor that switches between normal and provisioning mode.
In normal mode nuimo_app is running and Bluenet not while in provisioning mode Bluenet is started and nuimo_app is stopped.

This is necessary because the setup app cannot be connected while Nuimos are connected and the other way around.

### DIFF
--- a/senic_hub/backend/netwatch.py
+++ b/senic_hub/backend/netwatch.py
@@ -1,0 +1,135 @@
+#!/usr/bin/python3
+
+import click
+import dbus
+import dbus.mainloop.glib
+import logging
+import time
+import xmlrpc.client
+
+import NetworkManager
+try:
+    from gi.repository import GObject
+except ImportError:
+    import gobject as GObject
+
+from .supervisor import start_program, stop_program
+
+logger = logging.getLogger(__name__)
+
+
+@click.group()
+@click.pass_context
+@click.option('--wlan', '-w', required=False, help="WLAN device (default = wlan0)")
+def netwatch_cli(ctx, wlan):
+    if not wlan:
+        wlan = 'wlan0'
+    ctx.obj = NetwatchSupervisor(wlan)
+
+
+@netwatch_cli.command(name='start', help="start Netwatch supervisor")
+@click.option('--verbose', '-v', count=True, help="Print info messages (-vv for debug messages)")
+@click.pass_context
+def netwatch_start(ctx, verbose):
+    log_format = '%(asctime)s %(levelname)-5.5s [%(name)s] %(message)s'
+    if verbose >= 2:
+        logging.basicConfig(level=logging.DEBUG, format=log_format)
+    elif verbose >= 1:
+        logging.basicConfig(level=logging.INFO, format=log_format)
+    else:
+        logging.basicConfig(level=logging.WARNING, format=log_format)
+    ctx.obj.run()
+
+
+class NetwatchSupervisor(object):
+    """
+    Watches for changes in the network connection.
+    If the connection is down, Bluenet will be started to be able to use the setup app
+    to configure the Wifi connection and nuimo_app will be stopped because the setup app can
+    only connect if there are no Nuimos connected.
+    If the connection is back up, the nuimo_app will be restarted.
+    """
+
+    def __init__(self, wlan_adapter):
+        """
+        Initializes the object. It can then be started with run().
+        :param wlan_adapter: name of the WLAN adapter (i.e. 'wlan0')
+        """
+        self._wlan_adapter = wlan_adapter
+        self._bluenet_is_running = False
+        self._nuimo_app_is_running = False
+        self._bluenet_rpc = xmlrpc.client.ServerProxy('http://127.0.0.1:6459')
+
+    def run(self):
+        dbus.mainloop.glib.DBusGMainLoop(set_as_default=True)
+
+        # check initial status:
+        state = NetworkManager.NetworkManager.State
+        self._on_state_changed(None, state)
+
+        # listen for changes:
+        NetworkManager.NetworkManager.OnStateChanged(self._on_state_changed)
+        logger.debug("Start listening to network status changes")
+        loop = GObject.MainLoop()
+        loop.run()
+
+    def _on_state_changed(self, nm, state, **kwargs):
+        logger.info("State changed to %d: %s" % (state, NetworkManager.const('STATE', state)))
+        if state >= NetworkManager.NM_STATE_CONNECTED_GLOBAL:
+            self._switch_to_normal_mode()
+        elif state <= NetworkManager.NM_STATE_DISCONNECTED:
+            self._switch_to_provisioning_mode()
+
+    def _switch_to_normal_mode(self):
+        while (self._bluenet_is_connected() and
+                NetworkManager.NetworkManager.State >= NetworkManager.NM_STATE_CONNECTED_LOCAL):
+            logger.debug("Waiting before leaving provisioning mode till setup app is disconnected...")
+            time.sleep(5)
+
+        if NetworkManager.NetworkManager.State < NetworkManager.NM_STATE_CONNECTED_LOCAL:
+            logger.debug("Staying in provisioning mode because connection is lost again")
+            return
+
+        logger.debug("Normal mode: Stopping Bluenet and starting nuimo_app")
+        try:
+            stop_program('bluenet')
+        except xmlrpc.client.Fault as e:
+            logger.warning("Error while stopping Bluenet: %s" % str(e))
+
+        try:
+            start_program('nuimo_app')
+        except xmlrpc.client.Fault as e:
+            if e.faultCode == 60:
+                logger.debug("nuimo_app is already running")
+                pass
+            else:
+                logger.warning("Error while starting nuimo_app: %s" % str(e))
+
+    def _switch_to_provisioning_mode(self):
+        logger.debug("Provisioning mode:  Stopping nuimo_app and starting Bluenet")
+        try:
+            stop_program('nuimo_app')
+        except xmlrpc.client.Fault as e:
+            logger.warning("Error while stopping nuimo_app: %s" % str(e))
+
+        try:
+            start_program('bluenet')
+        except xmlrpc.client.Fault as e:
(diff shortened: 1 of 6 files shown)


senic_hub/backend/netwatch.py
```diff
@@ -0,0 +1,135 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+@click.group()
+@click.pass_context
+@click.option('--wlan', '-w', required=False, help="WLAN device (default = wlan0)")
+def netwatch_cli(ctx, wlan):
+    if not wlan:
+        wlan = 'wlan0'
```
larsblumberg: Alternative way of expressing this: `wlan = wlan or 'device'`, in case you want to choose which one you prefer

senic_hub/backend/netwatch.py
```diff
@@ -0,0 +1,135 @@
...
+    """
+
+    def __init__(self, wlan_adapter):
+        """
+        Initializes the object. It can then be started with run().
+        :param wlan_adapter: name of the WLAN adapter (i.e. 'wlan0')
+        """
+        self._wlan_adapter = wlan_adapter
+        self._bluenet_is_running = False
+        self._nuimo_app_is_running = False
```
larsblumberg: `_bluenet_is_running` and `_nuimo_app_is_running` are not used any longer

senic_hub/backend/netwatch.py
```diff
@@ -0,0 +1,135 @@
...
+
+        # check initial status:
+        state = NetworkManager.NetworkManager.State
+        self._on_state_changed(None, state)
+
+        # listen for changes:
+        NetworkManager.NetworkManager.OnStateChanged(self._on_state_changed)
+        logger.debug("Start listening to network status changes")
+        loop = GObject.MainLoop()
+        loop.run()
```
larsblumberg: Can the `netwatch` process be reliably stopped via `supervisor stop netwatch`? I'm not sure if that works given that the run loop is not stopped anywhere.

senic_hub/backend/netwatch.py
```diff
@@ -0,0 +1,135 @@
...
+        if state >= NetworkManager.NM_STATE_CONNECTED_GLOBAL:
+            self._switch_to_normal_mode()
+        elif state <= NetworkManager.NM_STATE_DISCONNECTED:
+            self._switch_to_provisioning_mode()
+
+    def _switch_to_normal_mode(self):
+        while (self._bluenet_is_connected() and
+                NetworkManager.NetworkManager.State >= NetworkManager.NM_STATE_CONNECTED_LOCAL):
+            logger.debug("Waiting before leaving provisioning mode till setup app is disconnected...")
+            time.sleep(5)
```
larsblumberg: Neat!

senic_hub/backend/netwatch.py
```diff
@@ -0,0 +1,135 @@
...
+
+    def _on_state_changed(self, nm, state, **kwargs):
+        logger.info("State changed to %d: %s" % (state, NetworkManager.const('STATE', state)))
+        if state >= NetworkManager.NM_STATE_CONNECTED_GLOBAL:
+            self._switch_to_normal_mode()
+        elif state <= NetworkManager.NM_STATE_DISCONNECTED:
+            self._switch_to_provisioning_mode()
+
+    def _switch_to_normal_mode(self):
+        while (self._bluenet_is_connected() and
```
larsblumberg: Since methods preferably start with a verb, this one could be named `is_bluenet_connected()`.

You could also make it a property, that's not necessary though IMHO.

senic_hub/backend/netwatch.py
```diff
@@ -0,0 +1,135 @@
...
+                NetworkManager.NetworkManager.State >= NetworkManager.NM_STATE_CONNECTED_LOCAL):
+            logger.debug("Waiting before leaving provisioning mode till setup app is disconnected...")
+            time.sleep(5)
+
+        if NetworkManager.NetworkManager.State < NetworkManager.NM_STATE_CONNECTED_LOCAL:
+            logger.debug("Staying in provisioning mode because connection is lost again")
+            return
+
+        logger.debug("Normal mode: Stopping Bluenet and starting nuimo_app")
```
larsblumberg: I like that most of your message start with a verb that say what net watch is going to do next. Here it netwatch would better tell about what it's doing right now if you change the log message to:

    logger.debug("Entering normal mode: Stopping Bluenet and starting Nuimo App")

I find it preferable if you use a consistent way of spelling: `Bluenet and Nuimo App` vs. `bluenet and nuimo_app`.

senic_hub/backend/netwatch.py
```diff
@@ -0,0 +1,135 @@
...
+            stop_program('bluenet')
+        except xmlrpc.client.Fault as e:
+            logger.warning("Error while stopping Bluenet: %s" % str(e))
+
+        try:
+            start_program('nuimo_app')
+        except xmlrpc.client.Fault as e:
+            if e.faultCode == 60:
+                logger.debug("nuimo_app is already running")
+                pass
```
larsblumberg: No need for `pass` here, right?

senic_hub/backend/netwatch.py
```diff
@@ -0,0 +1,135 @@
...
+            else:
+                logger.warning("Error while starting nuimo_app: %s" % str(e))
+
+    def _switch_to_provisioning_mode(self):
+        logger.debug("Provisioning mode:  Stopping nuimo_app and starting Bluenet")
```
larsblumberg: `Entering provisioning mode: Stopping ....`

senic_hub/backend/netwatch.py
```diff
@@ -0,0 +1,135 @@
...
+        state = NetworkManager.NetworkManager.State
+        self._on_state_changed(None, state)
+
+        # listen for changes:
+        NetworkManager.NetworkManager.OnStateChanged(self._on_state_changed)
+        logger.debug("Start listening to network status changes")
+        loop = GObject.MainLoop()
+        loop.run()
+
+    def _on_state_changed(self, nm, state, **kwargs):
```
larsblumberg: To make it more obvious, maybe name it `_on_network_state_changed`? Since we are also observing the phone connection state via BLE.

senic_hub/bluenet/bluenet.py
```diff
@@ -317,6 +323,19 @@ def _get_ip_address(self):
         ip = ifconfig_output.split('addr:', 1)[1].split(' ', 1)[0]
         return ip
 
+    def _start_rpc_server(self):
+        def bluenet_is_connected():
```
larsblumberg: To start methods with a verb, you can name it: `is_bluenet_connected`.

Exceptions from methods not starting with a verb can be event handlers such as `on_button_clicked`.

[I once started a simple style guide for naming](https://github.com/getsenic/coding-style), but we didn't actively maintain it. The facts however still apply.

senic_hub/backend/netwatch.py
luminosuslight: Supervisor sends a TERM signal ([see here under stopsignal](http://supervisord.org/configuration.html#program-x-section-settings)) to netwatch which would just stop the MainLoop as there are no threads involved (just tested this with a manual TERM signal). It is the same as with the Bluenet BlueZ peripheral, or did I miss something?

larsblumberg: Thanks for clarifying, that totally answers my question. It's all good then!

senic_hub/backend/netwatch.py
luminosuslight: The reason for `Normal mode:` was that it is possible that this method is executed multiple times in a row (i.e. because network state changed from `connecting` to `connected local` to `connected global`) and I didn't wanted `Entering normal mode:` to appear twice in the log. Without further logic it is not possible to check if it already is in a certain mode, so I would leave it like that (except for the capitalization) but I'm open for arguments.

senic_hub/backend/netwatch.py
luminosuslight: [see above](https://github.com/getsenic/senic-hub/pull/194#discussion_r124005959)

senic_hub/backend/netwatch.py
larsblumberg: I get your point – I'm totally fine with seeing the current message as is multiple times then.

senic_hub/bluenet/bluenet.py
luminosuslight: I get a 404 error with the link :-/

larsblumberg: It's still a private repo, when you log in, you should see it

larsblumberg: My bad – it wasn't even in our group. Now you should see it.

luminosuslight:
I will wait with the merge till I had the chance to test it on a NanoPi because this is a critical component.

larsblumberg:
@luminosuslight I can test it here. I would simply run it, constantly connect and disconnect it from wifi by setting/deleting the wifi connection and then check if the hub is discoverable via BLE when it's not connected. That should be sufficient, no?

luminosuslight:
Yes, I already tested it so only typos are expected. Another thing to check is of course if the hub is **not** discoverable when the Wifi is connected.

senic_hub/backend/netwatch.py
```diff
@@ -0,0 +1,131 @@
...
+    If the connection is back up, the nuimo_app will be restarted.
+    """
+
+    def __init__(self, wlan_adapter):
+        """
+        Initializes the object. It can then be started with run().
+        :param wlan_adapter: name of the WLAN adapter (i.e. 'wlan0')
+        """
+        self._wlan_adapter = wlan_adapter
+        self._bluenet_rpc = xmlrpc.client.ServerProxy('http://127.0.0.1:6459')
```
tomster: IMHO a good example where it's better to simply hardcode something (that's under our control, anyway) instead of automatically reaching for yet another configurable option.

luminosuslight:
https://github.com/getsenic/senic-os/issues/12 needs to be done in order for this PR to work, too.

luminosuslight:
This PR is included in #218.

Issue #199, opened 2017-06-26
LIFX brightness 0 issue


commento:
The LIFX component perceives brightness 0 as a valid brightness.
If the brightness value is decreased till 0 with nuimo and the light is then turned-on after this, the light stays off. 
The unexpected behaviour is also that sometimes the light turn on even if brightness 0 is applied.
To be analyzed if the issue is in nuimo app or lifx home assistant component but the second option is more probable because it does not takes place with lightify and dali (light nuimo app component) through home assistant.
```
2017-06-26 12:59:24,060 DEBUG nuimo_app.components.light Is On: False, Brightness: None
2017-06-26 12:59:24,964 DEBUG nuimo_app Ignoring gesture event: Gesture.BUTTON_RELEASE
2017-06-26 13:00:33,697 DEBUG nuimo_app Processing gesture event: Gesture.BUTTON_PRESS
2017-06-26 13:00:33,700 DEBUG nuimo_app.components Call service light:turn_on with data: {'entity_id': 'light.lifx_bulb_203d10'}
2017-06-26 13:00:33,701 DEBUG nuimo_app.hass Sending request id: 378
2017-06-26 13:00:33,702 DEBUG nuimo_app.hass {'domain': 'light',
 'id': 378,
 'service': 'turn_on',
 'service_data': {'entity_id': 'light.lifx_bulb_203d10'},
 'type': 'call_service'}
2017-06-26 13:00:33,731 DEBUG nuimo_app.hass Received result id: 378 success: True result:
2017-06-26 13:00:33,732 DEBUG nuimo_app.hass None
2017-06-26 13:00:33,733 DEBUG nuimo_app.components Calling service light:turn_on succeeded with result: {'id': 378, 'result': None, 'success': True, 'type': 'result'}
2017-06-26 13:00:33,757 DEBUG nuimo_app.components.light Updated light state: {'attributes': {'brightness': 0,
                'color_temp': 400.0,
                'friendly_name': 'LIFX Bulb 203d10',
                'max_mireds': 500,
                'min_mireds': 154,
                'rgb_color': [0, 0, 0],
                'supported_features': 51},
 'entity_id': 'light.lifx_bulb_203d10',
 'last_changed': '2017-06-26T13:00:33.738737+00:00',
 'last_updated': '2017-06-26T13:00:33.738737+00:00',
 'state': 'on'}
2017-06-26 13:00:33,758 DEBUG nuimo_app.components.light Is On: True, Brightness: 0
[...]

PR #201, opened 2017-06-26
Added turn-off command when brightness is 0

commento:
in case of brightness value 0 reached with nuimo rotation.

This fix the LIFX brightness 0 issue #199 and does not introduce drawbacks (tested with Dali light and lightify components through HA)


larsblumberg:
Very good!

larsblumberg:
Please squash + merge

tomster:
Looks good, but at some point we really need to add tests for this.
A workable test setup is not trivial, though, as we need to find a fine balance between mocking too much and mocking too little...
But in the end, this is just me mumbling "we need tests. you fools!" without adding anything substantial to the issue :-)

PR #208, opened 2017-07-03
Fix a wrong logging command

luminosuslight:
Probably the line was changed from print() to logger.info(), but results in an exception the way it was.


senic_hub/nuimo_app/__init__.py
```diff
@@ -65,7 +65,7 @@ def __init__(self, ha_api_url, ble_adapter_name, mac_address, components):
 
     def start(self):
         self.set_active_component()
-        logger.info("Connecting to Nuimo controller", self.controller.mac_address)
+        logger.info("Connecting to Nuimo controller %s" % self.controller.mac_address)
```
larsblumberg: I learned from @grunskis that the preferred way to log with placeholders is:

because it would not evaluate all the arguments on the right side unless the log message is actually printed to the log (which it might not due to a restrictive log levels)

luminosuslight: Ah ok, changed it and will think of it in the future :+1:

PR #215, opened 2017-07-07
Add workaround to fix Android 5.1 bug

### LINKED ISSUE #170: Bluenet BLE service not found on devices running Android 5.1
The custom Bluetooth Low Energy GATT service that Bluenet is advertising for Wifi provisioning is not showing up on multiple devices running Android 5.1. Instead two default services are shown.

This can be observed in both the React Native app and Nordic Semiconductor nRF Connect app, so it is probably not a problem in the app itself.
Bluetooth LE and 128 bit UUIDs in general can be eliminated as the possible issue, too, as these are working on at least one of these devices (Nuimos can be connected).
The output of logcat is identical to successfully connecting to a Nuimo.

Known problematic devices:
* Huawei Mate S running Android 5.1.1
* @grunskis phone? - Fairphone 2 with Android 5.1

Unsolved Stackoverflow question with the same problem: [Here](https://stackoverflow.com/questions/38002948/android-ble-custom-service-uuid-not-finding)

luminosuslight:
This is a workaround that fixes #170.
Android 5.1 (and probably earlier versions) try to connect with BR/EDR to the hub which doesn't work. This PR disables BR/EDR on the hub to force a Bluetooth LE connection.

Eventually it would be good to either add a fix in the Android app or move this setting to a configuartion file on the hub (i.e. in the [main.conf settings file of BlueZ](https://github.com/RadiusNetworks/bluez/blob/master/src/main.conf#L53)).


senic_hub/bluenet/bluez_peripheral.py
```diff
@@ -274,6 +277,18 @@ def _unregister(self):
...
         logger.info("Peripheral unregistered")
 
+    def _disable_br_edr(self):
+        # Android 5.1 tries to connect with BR/EDR and not with BLE
+        # this is a workaround to prevent this
+        logger.info("Disabling BR/EDR...")
+        was_powered = self.is_powered
+        self.is_powered = False
+        try:
+            call(['btmgmt', 'bredr', 'off'])
```
larsblumberg: Does the run user have enough permissions to execute the command?

luminosuslight: Added a TODO for this, currently `user={{ run_user }}` is commented out in bluenet_supervisor.conf.

PR #219, opened 2017-07-11
Add hostname suffix

rudolfs:
In order to debug beta tester devices, we need a way to uniquely identify them.
The ploy_hostname which we define in ploy.conf is now used for the suffix.

  [plain-instance:duncan] sets the hostname to senic-hub-duncan


deployment/roles/hub-ubuntu/tasks/main.yml
```diff
@@ -28,3 +28,13 @@
 
 - include: ntp.yml
   tags: time
+
+- name: configure hostname
+  copy:
+    content: "senic-hub-{{ploy_hostname}}"
```
larsblumberg: If we rename `ploy_hostname` to `ploy_hostname_suffix` it would better expressive that its append at the end of the host name, `ploy_hostname` doesn't contain the full host name. What do you think?

rudolfs: It's an internal variable of ploy, not sure how to change that. Will leave it for now as is.

PR #226, opened 2017-07-14
Clean up setup API endpoints and remove unused code

larsblumberg:


tomster:
+1 FTR

larsblumberg:
As part of this PR, `senic_hub/backend/tests/test_setup_config.py` should have been renamed, too, namely to `setup_config.py`. Doing so will however lead to failing tests as renaming that file will change the order of the tests being executed. We have identified that various files from `tests/data/` are not copied before being modified. This leads to consecutive tests not getting the original content of these files, thus they fail. With a subsequent PR we will address this issue, probably by commonly sharing the modified application settings that will link to the copies of the original data files for each test case. @tomster Let's do that together today.

PR #233, opened 2017-07-21
Watch nuimo_app config file for changes

luminosuslight:
**Important:**  pyinotify needs to be installed for this PR to work.
It should only be merged when this is done in Senic OS. (see https://github.com/getsenic/senic-os/issues/26)

~**TODO:** It currently watches the whole data folder for changes, check why it doesn't work with a single file.~ fixed


larsblumberg:
I added an issue for this PR: #236

senic_hub/nuimo_app/__init__.py
```diff
@@ -196,3 +194,18 @@ def show_error_matrix(self):
 
     def display_matrix(self, matrix, **kwargs):
         self.controller.display_matrix(LedMatrix(matrix), **kwargs)
+
+    def set_components(self, components):
```
larsblumberg: This method is very important to `Nuimo_App`:
- it's the first method being called in its constructor
- it's being called before `start()`
- it's part of the public API of `Nuimo_App`,

hence it should go as the first method after the c'tor, not at the end

senic_hub/nuimo_app/__init__.py
```diff
@@ -196,3 +194,18 @@ def show_error_matrix(self):
...
+        # TODO: check if components is empty?
+
+        for component in components:
+            component.nuimo = self
+
+        self.components = components
+
+        if self.active_component:
+            self.active_component.stop()
```
larsblumberg: We should stop the previously active component before we assign the new components. Otherwise we might have an active component that's not part of the new components. That's a conflict and can lead to unwanted behavior.

I suggest this order:

- Stop previously active component and remember its ID
- Assign new components
- Make previously active component active again if still present

luminosuslight:
I addressed the requested changes and for me it works now as it should.
I also added some TODOs for how to make it support multiple Nuimos.

Cannot be merged yet because pyinotify is still missing in Senic OS.

commento:
Tested with my setup and pip-installing pyinotify. It works for me and it could be merged when pyinotify is added to senic-os

PR #234, opened 2017-07-21
Support multiple nuimos in nuimo config

commento:
handling based on mac_address


commento:
nuimo_app changes missing, to be completed

PR #240, opened 2017-07-28
Add Readme file for Bluenet

luminosuslight:
Adds a README file for Bluenet and the BlueZ peripheral library.

Ideally, together with the documentation in the code it should be possible to understand, use and continue develop Bluenet after reading it.


tomster:
Very nice. The test failures are not caused by this PR so i'm merging it despite the failed checks

PR #243, opened 2017-08-03
Add threading for multiple nuimos handling

commento:


senic_hub/nuimo_app/__init__.py
```diff
@@ -80,7 +80,11 @@ def start(self):
         self.set_active_component()
         logger.info("Connecting to Nuimo controller %s", self.controller.mac_address)
         self.controller.connect()
-        self.manager.run()
+        try:
+            self.manager.run()
+        except KeyboardInterrupt:
+            logger.debug("Stopping... Nuimo controller %s", self.controller.mac_address)
+            self.stop()
```
rudolfs: Is the exception handling for handling a user pressing CTRL-C, or something like that? Would that ever happen if it's run as a thread of a daemon? Or am I off the track here?

commento: this interrupt takes place also when the nuimo_app is stopped with supervisorctl

senic_hub/nuimo_app/__main__.py
```diff
@@ -35,31 +35,30 @@ def main(config):
         config = yaml.load(f)
 
     nuimos = config['nuimos']
-    nuimo_controller_mac_address = list(nuimos.keys())[0]
 
-    if not nuimo_controller_mac_address:
+    nuimo_controllers_mac_address = list(nuimos.keys())
```
commento: ok!

senic_hub/nuimo_app/__init__.py
rudolfs: Gotcha!

rudolfs: We probably should use consistent logging levels though. The previous logger goes into info. This one should also probably go on the same level.

senic_hub/nuimo_app/__main__.py
```diff
@@ -35,31 +35,30 @@ def main(config):
...
+
     if platform.system() == 'Linux':
         watch_config_thread = Thread(
             target=watch_config_changes, args=(config_file_path, nuimo_apps), daemon=True)
         watch_config_thread.start()
 
-    try:
-        nuimo_app.start()
-    except KeyboardInterrupt:
-        logger.debug("Stopping...")
```
commento: @rudolfs I took the logging level from here

rudolfs: I know, but it makes sense to have it consistent. In the other file we use info, I guess you can choose which one you prefer, but they should be the same, no?

commento: it is the same exception and the same trace moved from here to `__init__.py` that's why I used debug also there. Anyway I can change it to info

senic_hub/nuimo_app/__main__.py
```diff
@@ -35,31 +35,30 @@ def main(config):
...
     ble_adapter_name = app.registry.settings['bluetooth_adapter_name']
-    components = nuimos[nuimo_controller_mac_address]['components']
-    component_instances = get_component_instances(components)
-    nuimo_app = NuimoApp(ha_api_url, ble_adapter_name, nuimo_controller_mac_address, component_instances)
-    # TODO: create multiple NuimoApps for all Nuimos mentioned in the config file
 
     nuimo_apps = {}
-    nuimo_apps[nuimo_controller_mac_address] = nuimo_app
+
+    for nuimo_controller_mac_address in nuimo_controllers_mac_address:
```
commento: 2c085d94333d27b7ea86931e88150f8b4bd28014

senic_hub/nuimo_app/__main__.py
commento: 2c085d94333d27b7ea86931e88150f8b4bd28014

senic_hub/nuimo_app/__main__.py
```diff
@@ -107,6 +106,7 @@ def reload_config_file(config_path, nuimo_apps):
     for mac_addr in nuimo_apps.keys():
         if mac_addr not in nuimos:
             logger.info("A Nuimo was removed from the config file: %s", mac_addr)
+            nuimo_apps[mac_addr].stop()
             # TODO: stop NuimoApp for deleted Nuimo here
```
commento: 28a1f23113f552102fb4ef5e0365c9d090aa208a

senic_hub/nuimo_app/__init__.py
commento: 8aab544baa3bfd26142b194d91be7aee1464cb74

commento:
The code has been tested and does not introduce regression errors. The code presents the issue #244 still under investigation but does not introduce errors on the normal functioning. 
It has been decided to merge this PR even if this issue is present so that, in the meantime, it is possible to adapt the setup_app for multiple nuimos and test it on the master branch.

Issue #245, opened 2017-08-03
nuimo_app: multiple nuimos interdependence 


commento:
This issue concerns the functioning of multiple nuimos when connected. In case one of the two nuimos is turned off or it disconnects then the other is not working till the disconnected nuimos reconnect to the nuimo_app

luminosuslight:
Fixed in #250.

PR #250, opened 2017-08-09
Create nuimo_app process for each Nuimo

luminosuslight:
Adds support for multiple Nuimos and to add them at runtime. Starts a process for each one because gatt-python seems not to handle multiple devices in a single thread correctly.

This is still very verbose for debugging and many debug messages can be removed, but I leave them in for now to make the code more readable.


commento:
I reviewed and tested it. It works as expected.

commento:
it solves issue #245

Issue #262, opened 2017-08-22
HUE bridge connection error on NUIMO APP


commento:
Error occurring often when Philips hue component is active. 
Correct behaviour is restored moving to another component and returning to the Philips hue component.
To be handled correctly.
`2017-08-04 08:58:51,066 ERROR [senic_hub.nuimo_app.components] [Errno 104] Connection reset by peer
Traceback (most recent call last):
  File "/home/root/senic-hub/senic_hub/nuimo_app/components/__init__.py", line 53, in _run
    self.run()
  File "/home/root/senic-hub/senic_hub/nuimo_app/components/philips_hue.py", line 330, in run
    self.lights.update_state()
  File "/home/root/senic-hub/senic_hub/nuimo_app/components/philips_hue.py", line 162, in update_state
    response = self.bridge.get_light()
  File "/usr/lib/python3.5/site-packages/phue.py", line 846, in get_light
    return self.request('GET', '/api/' + self.username + '/lights/')
  File "/usr/lib/python3.5/site-packages/phue.py", line 663, in request
    result = connection.getresponse()
  File "/usr/lib/python3.5/site-packages/raven/breadcrumbs.py", line 346, in getresponse
    rv = real_getresponse(self, *args, **kwargs)
  File "/usr/lib/python3.5/http/client.py", line 1197, in getresponse
    response.begin()
  File "/usr/lib/python3.5/http/client.py", line 297, in begin
    version, status, reason = self._read_status()
  File "/usr/lib/python3.5/http/client.py", line 258, in _read_status
    line = str(self.fp.readline(_MAXLINE + 1), "iso-8859-1")
  File "/usr/lib/python3.5/socket.py", line 575, in readinto
    return self._sock.recv_into(b)
ConnectionResetError: [Errno 104] Connection reset by peer`

commento:
as reported in #263 the connection loss issue has to be handled properly for both philips hue and sonos components in order to cover all the edge cases. This specific issue (reported above) has been fixed with the workaround merged in the related PR.

PR #263, opened 2017-08-22
Handle ConnectionResetError in nuimo_app for Philips Hue light

commento:
fix issue #262 
The exception is handled and does not throw a Traceback.
After the error occurrence, the light is controllable and the error matrix is not displayed.


senic_hub/nuimo_app/components/philips_hue.py
```diff
@@ -240,7 +240,10 @@ def __init__(self, component_config):
         seed()
 
     def create_lights(self, light_ids):
-        reachable_lights = self.filter_reachable(light_ids)
+        try:
+            reachable_lights = self.filter_reachable(light_ids)
+        except ConnectionResetError:
+            logger.error("Hue Bridge not reachable, handle exception")
         if not reachable_lights:
```
rudolfs: This will just throw an UnboundLocalError exception, because reachable_lights is never assigned when the exception happens...

commento: ok let's discuss it

commento: cf074d25919c8f30a5c7f1c73a3e8718ff69b4d7

commento:
as discussed the TODO comment for the library wrapper has been added. An issue will be opened to face the connection losses in both philips hue and sonos components

PR #279, opened 2017-08-29
Add groups incremental name to identify one from another

commento:
every time a group is created the name is changed adding a suffix with an incremental number based on the groups already present per device


senic_hub/backend/views/nuimo_components.py
```diff
@@ -105,6 +105,11 @@ def add_nuimo_component_view(request):
         except KeyError:
             return HTTPNotFound("No Nuimo with such ID")
 
+        group_number = [comp['name'] for comp in nuimo['components'] if " Group " in comp['name'] and component['ip_address'] in comp['name']]
```
rudolfs: Would be nice if you could split it up and make it a bit more readable.

commento: this is the more readable version:
```
        group_number = []
        for comp in nuimo['components']:
            if " Group " in comp['name'] and component['ip_address'] in comp['name']:
                group_number.append(comp['name'])
```
but in my opinion is better to leave the compact form, otherwise would requires more code for the pytest coverage

rudolfs: This is so much better! The unit tests should not change if the implementation is the same.

commento: the nested if condition adds a new line of code not tested `group_number.append(comp['name'])` while the previous implementation is just a single line of code and it is covered by the already existing test cases (see travis coverage)
schermata 2017-08-30 alle 11 55 33

PR #290, opened 2017-09-01
Fix nuimo_app network disconnection issue

### LINKED ISSUE #259: Nuimo app unresponsive if device is unreachable
The Nuimo device becomes visually unresponsive(& blank) when nuimo_app attempts connection to a device over the network and the device is unreachable.

neelsenic:
Closes #259

Implemented a method which sends a ping to the active component at every gesture event, so as to keep track of its reachability. This ping-check can monitor only those components whose IP addresses are present in the config. file, and which allow ICMP requests (responding to a ping).


senic_hub/nuimo_app/__init__.py
```diff
@@ -259,6 +272,11 @@ def listen_to_ipc_queue(self, ipc_queue):
                 self.stop()
                 return
 
+    def check_connection_on_gesture(self, host_ip):
+        param = "-c 1 -w 1"
+        status = (system_call("ping " + param + " " + host_ip) == 0)
+        return status
+
```
rudolfs: This method should be defined in every component class (the implementation would vary from component to component), best to name it `is_device_responsive` or something like that.

You could check the connection by getting the HTTP headers because most of the services work via HTTP. It might be that the device answers ICMP requests, but HTTP is down, then your approach won't work. Try something like this:

```
import httplib
connection = httplib.HTTPConnection("192.168.1.97", 80, timeout=1)

try:
  connection.request("HEAD", "/")
  response = connection.getresponse()
  True
except:
  False
```

Another improvement would be to throttle the checks, for example if the device is down, return False for the next 30sec.

senic_hub/nuimo_app/__init__.py
```diff
@@ -138,7 +140,18 @@ def process_gesture_event(self, event):
             self.show_error_matrix()
             return
 
-        self.process_gesture(event.gesture, event.value)
+        if self.active_component.ip_address is not None:
+            ip_addr = self.active_component.ip_address
+            if self.check_connection_on_gesture(ip_addr) is True:
+                self.process_gesture(event.gesture, event.value)
+                return
```
commento: maybe here is better to remove this handling and keep the other `self.process_gesture(event.gesture, event.value)` 

> line 154

 for both with-ip-address-ping-successful and without-ip-address scenarios (keeping only the else condition).
In that case the comment at line 153 can be removed.

senic_hub/nuimo_app/components/__init__.py
```diff
@@ -17,6 +17,13 @@ class BaseComponent:
 
     def __init__(self, component_config):
         self.component_id = component_config['id']
+        self.ip_address = None
+        try:
+            if component_config['ip_address'] is not None:
+                self.ip_address = component_config['ip_address']
+        except KeyError:
```
commento: here can be used the compact form:
`component_config.get('ip_address', None)`
instead of using the try-except block

senic_hub/nuimo_app/components/__init__.py
```diff
@@ -17,6 +17,13 @@ class BaseComponent:
...
+            pass
+
```
rudolfs: How about:
```
if 'ip_address' in component_config:
  self.ip_address = component_config['ip_address']
else:
  self.ip_address = None
```
?

senic_hub/nuimo_app/__init__.py
```diff
@@ -138,7 +140,18 @@ def process_gesture_event(self, event):
...
+
```
rudolfs: Get rid of the newline.

senic_hub/nuimo_app/components/__init__.py
grunskis: Or maybe this: `self.ip_address = component_config.get('ip_address', None)`

neelsenic: Great! Thanks :smile:

senic_hub/nuimo_app/__init__.py
neelsenic: Thanks!

neelsenic: I've checked that unfortunately, philips hue & Sonos listen on ports 80 & 1400 respectively.

PR #311, opened 2017-09-21
Add Nuimo connection status

### LINKED ISSUE #310: Add/Upgrade endpoint to check the if a Nuimo is on/off
It would probably make the most sense to add this to the ``/-/confnuimos`` endpoint seeing as you are getting a list of all seen Nuimos.

However, if there is a significant cost to this, it would be find to break it out to it's own endpoint.

This will be used to display the status of a Nuimo to a user via the frontend.

commento:
Closes #310 

the endpoint `confnuimos` now contains the `is_connected` information related to the Nuimo connectivity status.
The configuration file is updated coherently.


neelsenic:
:cool:

kelada:
It's important to note that this addition does add about 300ms to the request. If it becomes an issue down the line it can be broken out into it's own endpoint.

👍

commento:
Yes, a TODO comment about adding a new endpoint for the battery level is present

PR #315, opened 2017-09-21
Implement API specs via cornice_swagger

### LINKED ISSUE #308: Add Hub API documentation via swagger-ui for pyramid
There is no Hub API documentation, there should be, use [pyramid_swagger](https://github.com/striglia/pyramid_swagger/) or something like it.

@tomster mentioned swagger-ui: [example](http://petstore.swagger.io/) it appears to do the job.

I'm not sure where the best place for this service to run would be, on the hub (only in dev builds)? or on api.senic.com/docs?

neelotpalnag:
Implemented generation of JSON-formatted API specs using `cornice_swagger`. The specs can be generated by sending a GET request to the backend, as  `/-/api_doc`. The output can be pretty-parsed at  parsed through an online JSON parser, or  []([](https://swagger.io/swagger-editor/)), or using any online JSON parser. 
Closes #308 

**NOTE:** These commits require the `cornice_swagger` import (use pip3 cornice_swagger) & addition of `hub_api_version` in the `development.ini` file, and a **cornice>=2.0.0* (preferably 2.3.0) & ``cornice-swagger`` to be added in required installations (in `/senic-hub/setup.py`).
**TODO:** (In subsequent commits) 
[1]. Add Schemas for some PUT requests & responses to GET requests.
[2]. Discuss on whether to continue using online parser for JSON specs, or run a node server to access the SwaggerUI pretty-parsed webpage for the same.


neelsenic:
ppwwppww
The swaggerUI parser looks something like this:

neelsenic:
screenshot from 2017-09-21 18-13-30
The SwaggerUI parser looks something like :+1:

senic_hub/backend/views/api_descriptions.py
```diff
@@ -0,0 +1,17 @@
+descriptions = {
+    'appinfo': 'Basic Senic Hub & Setup App information',
+    'configuration_service': 'Apply new/updated configurations for Nuimo/Hub ',
+    'devices_service': 'Smart Home Devices accessible by the Hub',
+    'nuimo_components_service': 'Info on components configured to a Nuimo',
+    'nuimo_component_service': 'Info on a component configured to a Nuimo',
+    'nuimo_device_test_service': 'Test individual components configured to a Nuimo',
+    'connected_nuimos': 'Nuimos connected to the Hub',
```
commento: I would say: Mac address of the Nuimo discovered during onboarding procedure

senic_hub/backend/views/api_descriptions.py
```diff
@@ -0,0 +1,17 @@
...
+    'configured_nuimos': 'Nuimos configured by the Hub',
```
commento: Nuimos and related components information

kelada:
This is a very good start, I think we need to make sure that upgrading the version of cornice doesn't have unintended consequences.

In addition each endpoint should provide an example output.

@tomster we talked about having a development spin of the senicos image, perhaps we could include node + swagger just in that build?

neelsenic:
@kelada : I have added the fixes required post-upgrading the cornice version. I'll watch out for any other consequences. I guess I've covered all the attributes where cornice is being used.

tomster:
@kelada regarding the consequences of upgrading cornice, i think we should be good, since we do have tests. sticking to the older version was just pure stasis, there's no hard reason to not upgrade. in fact, this is a good occasion to finally upgrade :)

for development i think it's fine to run nodejs to generate live swagger views but i'm against shipping the product with a running instance of nodejs

this also means that it's easy to add npm etc. for the development build by not adding it as a recipe on the yocto level but simply adding it to the development playbook which needs to be applied after booting the image anyway

PR #323, opened 2017-09-28
Add check-for-update endpoint

commento:
the `is_updated` flag is checking for updates in nuimo connection status and sonos configuration


kelada:
As the frontend polls this endpoint, if the backend sees something has changed it should send the response data of ``/-/confnuimos``

kelada:
When I'm in the office next, I'll test this and then merge it.

senic_hub/backend/views/nuimos.py
```diff
@@ -110,6 +118,34 @@ def get_configured_nuimos(request):  # pragma: no cover,
...
+    adapter_name = request.registry.settings.get('bluetooth_adapter_name', 'hci0')
+    manager = ControllerManager(adapter_name=adapter_name)
+    for mac_address in config['nuimos']:
+        # check Nuimo connection Status
+        controller = Controller(mac_address=mac_address, manager=manager)
+        is_updated = True if config['nuimos'][mac_address]['is_connected'] != controller.is_connected() else is_updated
+        if is_updated:
+            return get_configured_nuimos(request)
+        if controller.is_connected():
+            is_updated = True if config['nuimos'][mac_address]['battery_level'] != get_nuimo_battery_level(mac_address, manager) else is_updated
```
kelada: I tested the PR and the following changes need to be made:
- `get_nuimo_battery_level` only takes one arg, you are passing in two, this breaks the code
- `config['nuimos'][mac_address]['battery_level']` thorws a KeyError, because the `battery_level` is never written to the `devices.json` file
- also, what are the performance advantages of this PR over calling `/-/confnuimos`? rough numbers would be nice

commento: yes, this is referring to the get_nuimo_battery_level() implementation before merging of PR #338.
I will merge the latest master changes on this branch in order to have this code aligned.

PR #324, opened 2017-09-28
Add PHue test timeout

neelotpalnag:
Add timeout in `put` request to Hue bridge.


senic_hub/backend/views/nuimo_components.py
```diff
@@ -381,14 +381,15 @@ def test_blink_phue(component_ip, component_username, id):  # pragma: no cover
...
+        requests.put(request_url_put, data=param_low, timeout=1)
         time.sleep(0.5)
-        requests.put(request_url_put, data=param_default)
-        return 1
+        requests.put(request_url_put, data=param_default, timeout=1)
+        return True
 
     except:
+        logger.error("Error while testing PHue: ")
         return None
```
commento: sonos test returns True/False, can we set it coherently

senic_hub/backend/views/nuimo_components.py
```diff
@@ -381,14 +381,15 @@ def test_blink_phue(component_ip, component_username, id):  # pragma: no cover
...
         time.sleep(0.5)
-        requests.put(request_url_put, data=param_low)
+        requests.put(request_url_put, data=param_low, timeout=1)
         time.sleep(0.5)
-        requests.put(request_url_put, data=param_default)
-        return 1
+        requests.put(request_url_put, data=param_default, timeout=1)
+        return True
 
     except:
```
commento: can you add the exception that is thrown? (here and also above)

senic_hub/backend/views/nuimo_components.py
neelotpalnag: Ok. Sure. I missed that.

PR #331, opened 2017-09-29
Implement API documentation using cornice swagger

### LINKED ISSUE #308: Add Hub API documentation via swagger-ui for pyramid

neelotpalnag:
Implemented generation of JSON-formatted API specs using cornice_swagger. The specs can be generated by sending a GET request to the backend, as /-/api_doc. The output can be pretty-parsed at parsed through an online JSON parser, or using an online JSON parser.
Closes #308 

NOTE: These commits require the cornice_swagger import (use pip3 cornice_swagger) & addition of hub_api_version in the development.ini file, and a *cornice>=2.0.0 (preferably 2.3.0) & cornice-swagger to be added in required installations (in /senic-hub/setup.py) in the os-recipe


neelotpalnag:
Needs a `Squash and Merge` instead of a Rebase. 
So no worries.

tomster:
sorry for being late to this but i have two issues:

- what's the motivation of putting the actual descriptions into a separate module and referencing them? the whole point of decorators is to have the relevant information in place of what's being decorated. we have now added another layer of abstraction for readers of the source code. i strongly propose to put the descriptions inside the decorators. one main function of this sort of generated documentation is to be able to generate an overview easily and quickly,  anyway, so the advantage of having all descriptions in one place (inside the python module) becomes much weaker.

- i see we are now performing input validation on the decorated methods - if this is the case we also should have at least some exemplary tests that show that this is actually working, i.e. do we really get 400s with invalid data etc.

having said all that: great stuff, though, @neelotpalnag! :)

neelotpalnag:
Thanks for the inputs @tomster 
1. My choice of putting descriptions in a separate module was just to implement its appearance in the SwaggerUI-generated JSON. But unfortunately, the description doesn't reflect in the documentation anyway, because SwaggerUI doesn't support that feature. So, it's better to just comment the description. 
I agree it wasn't a smart move to abstract.

2. In case of badly formatted input, the `colander_body_validator` only allows filling a `request.error` type variable with the error, depending on the extent to which schemas have been defined. The reason for adding the `colander_body_validator` is that `cornice==2.3.0` requires a validator defined in the cornice-service decorator. It doesn't facilitate controlling the nature of 4xx response. For bad requests, we already have a `HTTPBadRequest` response in place. We can definitely output the `request.error` in the response to request, but it hasn't been implemented yet.

I hope this answers your queries. I'd love to have more feedback on what is the ideal thing to do now.

PR #333, opened 2017-10-04
Align bluetooth restart in nuimo_app to latest senic os changes

commento:
`/etc/init.d/bluetooth stop/start` is no longer present and the code in nuimo_app has been updated coherently.
This should solve the issue related to battery level PR #327. The error is present only since the latest senic-os update.
Anyway, it is convenient to put the battery level information in a separate endpoint


commento:
the main issue has been faced by the PR #337. 
The battery level feature is now under review, in PR #338 and https://github.com/getsenic/nuimo-linux-python/pull/34.
The approach used here for battery level feature implementation in the backend app caused unexpected behaviours controlling nuimo (two instances for the same nuimo) and for this reason, has been abandoned.

PR #343, opened 2017-10-10
Fix Philips Hue group/scene coherence

commento:
The coherence of lights-in-scene with lights-in-group is checked every time a change takes place


commento:
@kelada, I am investigating the phue and soco module mocking but, as discussed, this will be faced in a separate PR.

kelada:
@commento when that PR is made I will merge both of these in. I'm trying to prevent new features being added without tests.

PR #345, opened 2017-10-10
Reset Hue Brightness

neelotpalnag:
Re-implementing the previous reset-brightness implementation for hue bulbs. This small addition is all thats needed to reset brightness. However, every nuimo has it's own instance of the `nuimo_app` process. So one nuimo cannot remember what the brightness set by the other nuimo was. That's something that was implemented in the `philips_hue library`. We could have eliminated this issue if we had our own implementation of the PHue API. 
It's possible to remove the above drawback, but at the cost of a time-expensive operation of updating data from the hue bridge on every press on the nuimo. It looks hideous when its slow.

Will attempt to find a work-around. For now, this fix seems satisfactory than no reset.


commento:
schermata 2017-10-11 alle 11 17 46

commento:
this feature reintroduces the lag in the button controller.
every light in the group is turned on separately and sometimes the procedure is not completed as expected. To be discussed if the lag is acceptable.
Also, the brightness update received from the bridge is not handled correctly (even if the implementation is the one reverted in the PR #325 ) and after some seconds the issue reappears identically.

neelotpalnag:
@commento : Lets make a final decision on this. People defere's no initely need the reset feature. There's no escaping that. Shall we kill this issue by the end of the day?

commento:
Error raised during testing:
`
2017-10-11 10:09:36,868 ERROR [senic_hub.nuimo_app.components.philips_hue] Failed to set light attributes: [{'type': 201, 'address': '/lights/10/state/bri', 'description': 'parameter, bri, is not modifiable. Device is set to off.'}]
`

neelotpalnag:
Closing PR - Attempting to modify the phue library.

PR #350, opened 2017-10-12
Reimplement Brightness reset feature

neelotpalnag:
Re-implemented previous revert #325 by @commento, with minor changes.


neelotpalnag:
@commento . Already looking to fix that. Thanks

PR #354, opened 2017-10-16
Fix Makefile for write-os

neelotpalnag:
Fixed Makefile in development folder for `write-os-ubuntu` & `write-osx`.


TheMeaningfulEngineer:
:+1:

PR #356, opened 2017-10-18
Handle FileNotFound

neelotpalnag:
The commits to this PR will handle exceptions arising in the logging, so as to reduce unnecessary events in the Sentry Logging.


neelotpalnag:
Will be solved with more structured commits to another pull request. Closing.

Issue #382, opened 2017-11-21
If you remove two or more Nuimos quickly throws at http 500 error


kelada:
Removing two or more Nuimos via the mobile app in quick succession causes the backend to throw a 500 error.

This is only really noticed in the frontend when using __DEV__ build of the mobile app, but nonetheless this error should be handled correctly.

``senic_hub.log``
```
2017-11-21 17:01:20,422 ERROR [waitress] Exception when serving /-/nuimos/d2-5b-9c-a8-3f-e0
Traceback (most recent call last):
  File "/usr/lib/python3.5/site-packages/waitress/channel.py", line 338, in service
    task.service()
  File "/usr/lib/python3.5/site-packages/waitress/task.py", line 169, in service
    self.execute()
  File "/usr/lib/python3.5/site-packages/waitress/task.py", line 399, in execute
    app_iter = self.channel.server.application(env, start_response)
  File "/usr/lib/python3.5/site-packages/raven/middleware.py", line 98, in __call__
    iterable = self.application(environ, start_response)
  File "/usr/lib/python3.5/site-packages/pyramid/router.py", line 236, in __call__
    response = self.invoke_subrequest(request, use_tweens=True)
  File "/usr/lib/python3.5/site-packages/pyramid/router.py", line 211, in invoke_subrequest
    response = handle_request(request)
  File "/usr/lib/python3.5/site-packages/pyramid/tweens.py", line 51, in excview_tween
    request_iface=request_iface.combined
  File "/usr/lib/python3.5/site-packages/pyramid/view.py", line 547, in _call_view
    response = view_callable(context, request)
  File "/usr/lib/python3.5/site-packages/pyramid/viewderivers.py", line 413, in viewresult_to_response
    result = view(context, request)
  File "/usr/lib/python3.5/site-packages/pyramid/tweens.py", line 22, in excview_tween
    response = handler(request)
  File "/usr/lib/python3.5/site-packages/pyramid/router.py", line 158, in handle_request
    view_name
  File "/usr/lib/python3.5/site-packages/pyramid/view.py", line 547, in _call_view
    response = view_callable(context, request)
  File "/usr/lib/python3.5/site-packages/pyramid/config/views.py", line 182, in __call__
    return view(context, request)
  File "/usr/lib/python3.5/site-packages/pyramid/viewderivers.py", line 393, in attr_view
    return view(context, request)
[...]

PR #385, opened 2017-11-27
Add method to get process status

neelotpalnag:
Handles cases when Nuimos are removed rapidly, without allowing the nuimo_app to restart completely.
Added a method to get the status of the process.


neelotpalnag:
Fix for PR #382

kelada:
This doesn't appear to work. Removing Nuimos quickly one after the other doesn't disconnect them at all. Only on reboot do they disconnect.

If a solution can't be found on the backend, a short term fix can be made on the frontend.

neelotpalnag:
I'd request applying a frontend fix because it seems to be spitting out errors constantly in the backend. The errors have been handled with try-catch,  but a permanent solution couldn't be found ATM

kelada:
Please test again and merge in

neelotpalnag:
Tested. Closing.

PR #387, opened 2017-12-11
Append Room name for hue bulbs

neelotpalnag:
The code for extracting the Room name of every bulb is added in this PR. However nuimos.py isn't the best place to keep this code. However, the logic of this code is correct & works fine.

**Since this feature isn't a requirement of the UI specs, it will be used if required, in the future**


neelotpalnag:
The PR a non-breaking non-interfering code. Can be safely merged, but won't serve any purpose, or break the backend

PR #388, opened 2017-12-11
Remove Ping on gestures & X

neelotpalnag:
The PR removes the ping method which showed X whenever an error occurred due to unreachability to the lights/speakers, due to API call latencies, and spat out 3 lines of ping stats everytime a gesture was received. The X isn't a pleasant UI feature at all.


kelada:
Test and merge in

Issue #328, opened 2017-09-29
Pyramid has a `config.route_prefix` to prefix all routes


larsblumberg:
Currently route paths are manually prefixed through the `path()`/`service_path()` function. See example here:

https://github.com/getsenic/senic-hub/blob/231a900a2245619a46be82f61cd7dabd1c606a87/senic_hub/backend/views/nuimos.py#L22

The `path()` method prefixes all route paths with `/-/`:

https://github.com/getsenic/senic-hub/blob/231a900a2245619a46be82f61cd7dabd1c606a87/senic_hub/backend/config.py#L17

[Pyramid has a configuration setting that can implicitly prefix all route paths](https://docs.pylonsproject.org/projects/pyramid/en/latest/api/config.html) that would would make the use of the `path()` method unnecessary:

```python
config = ...
config.route_prefix = '/-/'
```

neelotpalnag:
Implemented in #390

PR #390, opened 2017-12-20
Remove service_path() from endpoints

neelotpalnag:
Implements task #328


senic_hub/backend/config.py
```diff
@@ -16,10 +16,10 @@ def path(service):
     """ Return path — or route pattern — for the given REST service. """
     return '/-/{0}'.format(service.lower())
 
-
```
larsblumberg: Doesn't PEP8 fail here? It requires 2 empty lines between global methods.

neelotpalnag: True. I did run tox, but must have put a space while scrolling. I'll fix it asap. I also need to remove the unused `path()` method. Thanks for mentioning!

PR #391, opened 2017-12-22
Re-assign different levels to logs

neelotpalnag:
Following the discussion regarding QA, I've made some changes in what logs will be shown at INFO level, so as to ease the QA process. While some events have been moved from DEBUG to INFO, most have been removed from INFO level to other levels. 

This is supposed to give a better and concise picture of what's going on in different processes.


TheMeaningfulEngineer:
Took a different path to implementing this with @avichalp .
Closing.

PR #393, opened 2018-01-12
get network state from 3rd arg in  if it in not provided in 2nd arg

avichalp:
Grab `state` key from `kwargs` because it is not passed by in the 2nd argument by `NetworkManager V1.4`


TheMeaningfulEngineer:
Confirmed the commit fixes the wifi onboarding.
Merging and updating the os

PR #396, opened 2018-01-17
Improve pre wifi logs for Q/A

avichalp:
Add 3 log events for:
- Hub is ready for on boarding
- App is discovered/connected to hub
- Hub Connected to the wifi


TheMeaningfulEngineer:
There seems to be a problem with CI.
Could you pls check it out?

PR #397, opened 2018-01-22
fix device discovery logger + add more logs

avichalp:
The device discovery module is using pyramid's `setup_logging` method to configure logging.
Due to which current logs doesn't work because they log using `logging.getLogger`.
This PR also adds a couple of logs about Phue authentication.


avichalp:
@TheMeaningfulEngineer : Make sense, changing to that.

PR #401, opened 2018-01-26
Sentry logs

avichalp:
This PR configure the logs using `senic-hub.ini` file.


senic_hub/backend/device_discovery.py
```diff
@@ -22,6 +24,12 @@
 from .network_discovery import NetworkDiscovery
 
 
+if os.path.isfile('/etc/senic_hub.ini'):  # pragma: no cover
```
TheMeaningfulEngineer: What does existence of `/etc/senic_hub.ini` means behavior wise?

avichalp: If it is not present the loggers will have default configuration if it is there loggers will be configured using the settings it contains for the loggers.

TheMeaningfulEngineer: > if it is there loggers will be configured using the settings it contains for the loggers.
Why do we want that?

avichalp: Actually, we are sure that on the hardware the file will be present.
But for someone who is running tests on local machine or when tests are running on Travis the file is not present in those environments.

PR #402, opened 2018-01-29
fix path of logging conf

avichalp:
Fixes the path of `conf` file (senic-hub.ini).


senic_hub/nuimo_app/__main__.py
```diff
@@ -16,15 +16,16 @@
...
-from . import NuimoApp
-
 
 if os.path.isfile('/etc/senic_hub.ini'):
     logging.config.fileConfig(
         '/etc/senic_hub.ini', disable_existing_loggers=False
     )
 
 
+from . import NuimoApp
```
TheMeaningfulEngineer: Why is the import moved?

avichalp: I found out that logger inside the `component` module in `nuimo_app` was not configured.
I think the reason is because import is done before configuration.
I haven't got a chance to test this hypothesis. Will test and update here.
Also, may be we can take it up in a different PR? and I will just remove this line from here for now?

TheMeaningfulEngineer: Sure

PR #405, opened 2018-02-01
Add version, ssid and hw id to appinfo API endpoint

rudolfs:
These are needed in the "More" -> "Settings" screen of the app.
Remove unused senic_hub egg version and hub onboarding logic.

Relevant to: https://github.com/getsenic/senic-home-automation-app/issues/38


senic_hub/backend/views/appinfo.py
```diff
@@ -20,18 +20,40 @@
...
+def os_version():
+    osrelease = '/etc/os-release'
+
+    if not os.path.isfile(osrelease):
+        return('')
+
+    for line in open(osrelease):
+        if 'VERSION=' in line:
+            version = re.split('VERSION=', line)[1].strip().replace('"', '')
+            return(version)
```
avichalp: Nit: space between `return` and `(version)`. And at all other places.

Also, in python if you have one element in tuple it ends up like this:
```
>>> a = (1)
>>> a
1
```

So good to have a comma after single element.
```
>>> a = (1, )
>>> a
(1,)
```

rudolfs: I'm not sure I get the part about the tuple. The return value should always be a string. Let's ignore the fact that there might be an exception in case of missing "VERSION=" substring for now...

senic_hub/backend/views/appinfo.py
```diff
@@ -20,18 +19,49 @@
...
+    if not os.path.isfile(cpuinfo):
+        return ''
 
-    return (exists(nuimo_app_config_path) and
-            exists(devices_path) and
-            exists(homeassistant_config_path) and
-            exists(nuimo_mac_address_filepath))
+    for line in open(cpuinfo):
+        if 'Serial' in line:
+            serial = re.split(':\s*', line)[1].strip().replace('02c00081', '')
```
kelada: How do you know that this will always be present? Is it not just safer to display it all?

rudolfs: I do it because otherwise the info screen looks cluttered in the app.

The prefix is only removed if it's present, otherwise it will leave the id unchanged, so we don't really loose any information. The "02c00081" part is always the same for Nanopi NEOs:

http://linux-sunxi.org/SID_Register_Guide#Currently_known_SID.27s
https://docs.google.com/document/d/12xCqVusy0cDe4xoCGC1Dob_T8mvHbED_vw9SCfYeBq8/edit#

senic_hub/backend/views/appinfo.py
```diff
@@ -20,18 +19,49 @@
...
     return dict(
-        version=get_distribution('senic_hub').version,
-        onboarded=is_hub_onboarded(request)
+        wifi=wifi(),
+        os_version=os_version(),
+        hardware_identifier=hardware_identifier(),
     )
 
 
-def is_hub_onboarded(request):
```
kelada: Was anything else using this?

avichalp: No, this is just used here for `app_info` related `http` endpoint.

kelada: 👍

PR #410, opened 2018-02-02
change production logs to warning

avichalp:
Changes production level to warning.

Note: this doesn't effect the log level of `senic_hub` process (REST API). That will be addressed here: https://github.com/getsenic/senic-os/issues/87


TheMeaningfulEngineer:
Do you know why it isn't following the same pattern [here](https://github.com/getsenic/senic-hub/blob/prod-error-log-level/senic_hub/bluenet/bluenet.py#L48)

`logging` vs `logger` thing?
Does it make sense here to go for consistency throughout the app?

avichalp:
- Here in `bluenet` log configuration is done using `logging.basicConfig` where as in other apps log configuration is done using `logging.fileConfig` and `senic_hub.ini` file.

-  We do not want `bluenet` to use sentry (because sentry cannot work without wifi) at all and that is why we do not use the `senic-hub.ini` file to configure it.

It make sense to have consistency but I think requirement is a bit different here. I think a good practice to follow could be to use `logging` for configuration of logs and use specific `loggers` (eg: root logger) to log messages and override specific configurations if required (for eg: `-v` & `-vv`).

Thought?

TheMeaningfulEngineer:
Sure, makes sense.
So this line is just a leftover?
https://github.com/getsenic/senic-hub/blob/prod-error-log-level/senic_hub/bluenet/bluenet.py#L26

avichalp:
Looks like leftover. We haven't touched this part in this PR. But it is used for logging messages like `logger.info` and `logger.debug` etc: 
For eg: [here](https://github.com/getsenic/senic-hub/blob/prod-error-log-level/senic_hub/bluenet/bluenet.py#L130)

TheMeaningfulEngineer:
Ok, will merge it tomorrow. (Need to rebase everything on it  and it's getting late :) )

TheMeaningfulEngineer:
Merging this would cause to much of a mess on the BT branches.
Will apply in manually on to: https://github.com/getsenic/senic-hub/pull/411

TheMeaningfulEngineer:
Addressed here:
https://github.com/getsenic/senic-hub/pull/411/commits/571622d62f613230dda42e53ac4af0243816d672

Closing

Issue #407, opened 2018-02-02
Nuimo reconnect


TheMeaningfulEngineer:
Precondition: vanilla onboarding
Description:

After a onboarded first nuimo, go out of range
and reconnect when back in range.
The nuimo reconnects sometimes but at certain point
stops reconnecting.

TheMeaningfulEngineer:
Solved with https://github.com/getsenic/senic-hub/pull/424

Issue #406, opened 2018-02-02
Vanilla onboarding


TheMeaningfulEngineer:
Tracked https://github.com/getsenic/senic-hub/issues/394

Working branch: https://github.com/getsenic/senic-hub/tree/vanilla-onboarding

TheMeaningfulEngineer:
```
curl -X POST   http://10.10.10.159:6543/-/nuimos
```
produces an action that comes down to this method to connect to the physical nuimo.

TheMeaningfulEngineer:
Did some testing on timings:

Distance [m]: 3
Discovery time [s]: 0
Connection time [s]: 5
Total: 6

Covered hub
Distance [m]: 5.5
Discovery time [s]: 2s
Connection time [s]: timeout
Total: /

Covered hub
Distance [m]: 7
Discovery time [s]: 1s
Connection time [s]: 8 
Total: 9

Based on this short test taking the rough thresholds for timeouts:
Discovery: 3s
Connection: 10s

TheMeaningfulEngineer:
Solved with: https://github.com/getsenic/senic-hub/pull/411

Issue #408, opened 2018-02-02
Power board on off


TheMeaningfulEngineer:
Tracked https://github.com/getsenic/senic-hub/issues/394

Branch: https://github.com/getsenic/senic-hub/tree/power-on-of (based on [vanilla-onboarding](https://github.com/getsenic/senic-hub/tree/vanilla-onboarding))

TheMeaningfulEngineer:
Solved with: https://github.com/getsenic/senic-hub/pull/420

Issue #394, opened 2018-01-13
Fix bluetooth


TheMeaningfulEngineer:
[Hub branch](https://github.com/getsenic/senic-hub/tree/nrf52-onboarding-nuimo)
[Os branch](https://github.com/getsenic/senic-os/tree/porting-the-app-stack)

## Status

[Senic hub commit](https://github.com/getsenic/senic-hub/commit/cbab10241e11ddef9d4b83a01815bd45107e514f)
~~[Os commit](https://github.com/getsenic/senic-os/commit/8e974ca0e4a8ab8a73b40d994b9ab8de83fe1159)~~ OS is set to always take the latest commit of the senic-hub branch. For reproducibility you must manually set the commit in the recipe.

Nots: 
* Showing networks in the app might take some time (limit 20sec)
* Nuimo and peripherals detection can take some time (limit 20sec)

#### Solved by https://github.com/getsenic/senic-hub/pull/411 [** awaiting review **]

~~Random~~ success

- OK Connecting to wifi
- ~~Sometimes~~ OK Detect and connect first nuimo
- OK Detect Sonos and Hue and autenticate
- OK Reach screen after the video tutorial and control the compoenents with nuimo

When fails, the device can be discovered but not connected to:

	nuimoctl --discover --adapter hci0
	nuimoctl --adapter hci0 --connect e0:88:72:c4:49:c2

resterting bluetooth reanables discovery (and a nuimo in the app is shown):

	systemctl bluetooth restart
	nuimoctl --discover --adapter hci0
	nuimoctl --adapter hci0 --connect e0:88:72:c4:49:c2

#### Dev tip

For components that don't require authentication, this is a 
shortcut to the vanilla onboarding without the mobile app

	# Hub
	nmcli dev wifi con SSID  password PASSWORD  ifname wlan0

	# PC
	curl -X POST   http://10.10.10.159:6543/-/nuimos
	curl -X POST   http://10.10.10.159:6543/-/config
	curl -X GET   http://10.10.10.159:6543/-/confnuimos

### Nuimo reconnect https://github.com/getsenic/senic-hub/issues/407

Random success

Precondition: vanilla onboarding

- After a onboarded first nuimo, go out of range
and reconnect when back in range.

The nuimo reconnects sometimes but at certain point
stops reconnecting.

### Power board on off https://github.com/getsenic/senic-hub/issues/408

Precondition: vanilla onboarding

- Once the board is powered up the nuimo is 
reconnected and can control the devices

[...]

PR #411, opened 2018-02-04
Vanilla onboarding

TheMeaningfulEngineer:
This addresses https://github.com/getsenic/senic-hub/issues/406

The change history doesn't mean much of it's own so I tried to document and refactor methods 
to give you an overview on what's going on.
Please also check the commit messages (for more clarification).

Given that the issue of (sometimes) not being able to connect to a Nuimo during
onboarding was never traced to a cause, the only thing we can do is to assume
it was fixed by having it not show again (ever) during onboardings.

I onboarded 5+ times without problems, and tried some irregular combinations
(discover Nuimo but can't connect, can't discover, connect normally after previous two)
that didn't end up crashing the app. 

More than a confirmation if this works correctly, I would ask you to take a look if
the code is readable enough and give feedback on that.

I'm continuing on https://github.com/getsenic/senic-hub/issues/407,
based from this branch. Will merge this into nrf52-onboarding-nuimo once the 
review passes (please just review, don't merge it).

### DIFF
--- a/senic_hub/backend/commands.py
+++ b/senic_hub/backend/commands.py
@@ -20,12 +20,17 @@ import os
 import time
 
 
-def create_configuration_files_and_restart_apps(settings):
-    """ create_configuration_files_and_restart_apps
+def create_nuimo_app_cfg(settings):
+    """
+    Creates nuimo_app.cfg a hard dependency for nuimo_app
 
-    This is tightly coupled with the existence of
-    /data/senic-hub/devices.json
-    which probably somethin device_discovery creates
+    When there is no wifi present, netwatch kills the nuimo_app.
+    This methog responsible for the initial starting of the nuimo_app
+    once the hub got wifi connection the first time.
+    This happens during the hub onboarding.
+
+    Once the hub has been onboarded, the nuimo_app is started
+    by default by supervisor.
     """
 
     logger.info("Creating config files that map Nuimo to Sonos/Hue")
@@ -74,7 +79,6 @@ def create_configuration_files_and_restart_apps(settings):
             logger.debug("Writing %s into %s" % (config, nuimo_app_config_file_path))
             yaml.dump(config, f, default_flow_style=False)
 
-    # ALAN Netwatch isn't starting nuimo_app anymore
     if supervisor.program_status('nuimo_app') != 'RUNNING':
         supervisor.start_program('nuimo_app')
 
@@ -122,10 +126,6 @@ def phue_bridge_config(bridge):
 
 def generate_nuimo_app_configuration(nuimo_mac_address, devices):
     components = [create_component(d) for d in devices if d["authenticated"]]
-
-#    What are components?
-#    (Epdb) components
-#    [{'type': 'sonos', 'name': '10.10.10.114 - Sonos One', 'ip_address': '10.10.10.114', 'id': 'a6367f28-9363-445e-ab07-dd69f54f51ca', 'device_ids': ['7828ca17171e01400']}]
 
     config = {'nuimos': {}}
     config['nuimos'][nuimo_mac_address] = {
--- a/senic_hub/backend/commands.py
+++ b/senic_hub/backend/commands.py
@@ -16,13 +16,31 @@ COMPONENT_FOR_TYPE = {
 logger = logging.getLogger(__name__)
 
 import os.path
+import os
+import time
 
 
 def create_configuration_files_and_restart_apps(settings):
+    """ create_configuration_files_and_restart_apps
+
+    This is tightly coupled with the existence of
+    /data/senic-hub/devices.json
+    which probably somethin device_discovery creates
+    """
+
+    logger.info("Creating config files that map Nuimo to Sonos/Hue")
+
+    while not os.path.exists(settings['devices_path']):
+        logger.info("Waiting for Sonos or Hue to be detected")
+        logger.debug("Can't continue until %s has been created" %
+                     settings['devices_path'])
+        time.sleep(5)
+
     try:
         with open(settings['devices_path'], 'r') as f:
             devices = json.load(f)
-    except (FileNotFoundError, json.decoder.JSONDecodeError) as e:
+    except (json.decoder.JSONDecodeError) as e:
+        logger.error("%s doesn't contain a readable json" % settings['devices_path'])
         logger.error(e)
 
     nuimo_mac_address_file_path = settings['nuimo_mac_address_filepath']
--- a/senic_hub/backend/nuimo_setup.py
+++ b/senic_hub/backend/nuimo_setup.py
@@ -2,22 +2,80 @@ import logging
 import nuimo
 import gatt
 import threading
-
+import time
+import subprocess
 
 logger = logging.getLogger(__name__)
+
+# If any of these timeouts are reached,
+# both BlueZ and the bluetooth module
+# (via the reset pin) are reset
+# and the proces is retried
+DISCOVERY_TIMEOUT = 5
+CONNECTION_TIMEOUT = 10
+
+# Onboarding in the context of this file means
+# a succefull discovery and connecion to a nuimo
+MAX_ONBOARDING_ATTEMPTS = 3
 
 
 class NuimoSetup(nuimo.ControllerManagerListener, nuimo.ControllerListener):  # pragma: no cover
 
     def __init__(self, adapter_name):
-        self._gatt_manager = gatt.DeviceManager(adapter_name)
-        self._manager = nuimo.ControllerManager(adapter_name)
+        self._adapter_name = adapter_name
+        self._gatt_manager = gatt.DeviceManager(self._adapter_name)
+        self._manager = nuimo.ControllerManager(self._adapter_name)
         self._manager.listener = self
         self._is_running = False  # Prevents from considering D-Bus events if we aren't running
         self._discovery_timeout_timer = None
         self._connect_timeout_timer = None
         self._controller = None
         self._required_mac_address = None
+        self._discovery_duration = None
+        self._connect_duration = None
+        self._nuimo_connected = False
+        self._nuimo_onboarding_attempts = 0
+
+        # A control flag not to do any action while restart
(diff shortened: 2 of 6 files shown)


TheMeaningfulEngineer:
@avichalp Can you please review this so I can merge it.
Would be great that you try and see that you can make out what's happening here:
https://github.com/getsenic/senic-hub/pull/411/files#diff-fcecbd5461c7c1efc43b71ff1271ca7bL22

because that is where the whole Nuimo discovery starts. If something is unclear, please ask right away while the implementation is still fresh in my mind.

senic_hub/backend/commands.py
```diff
@@ -16,13 +16,31 @@
 logger = logging.getLogger(__name__)
 
 import os.path
+import os
+import time
 
 
 def create_configuration_files_and_restart_apps(settings):
+    """ create_configuration_files_and_restart_apps
```
avichalp: please change this to something like `creates config files and restarts the app` writing the name of the function again in the doc string doesn't make much sense.

senic_hub/backend/commands.py
```diff
@@ -16,13 +16,31 @@
...
+
+    This is tightly coupled with the existence of
+    /data/senic-hub/devices.json
+    which probably somethin device_discovery creates
```
avichalp: >which probably somethin device_discovery creates

Please make sure that the last line is true or not or just drop it. I think a having probabilistic comment is more confusing than useful for the person reading it later.  Thoughts?

senic_hub/backend/nuimo_setup.py
```diff
@@ -27,96 +85,230 @@ def discover_and_connect_controller(self, required_mac_address=None, timeout=Non
...
 
     def discovery_timed_out(self):
+        """
+        Callback. _start_discovery timed out.
+
+        Stops all bluetooth related actuivity and
+        initiates BT restart.
+        """
+
+        # Why is this false?
```
avichalp: Please remember to remove the "personal notes" type comments before merging.

senic_hub/backend/nuimo_setup.py
```diff
@@ -27,96 +85,230 @@ def discover_and_connect_controller(self, required_mac_address=None, timeout=Non
...
+        # This will trigger the discovery in terms of a d-bus command.
+        # https://github.com/getsenic/gatt-python/blob/master/gatt/gatt_linux.py#L139
         self._manager.start_discovery()
+        self._manager.run()
 
     def discovery_timed_out(self):
+        """
+        Callback. _start_discovery timed out.
+
+        Stops all bluetooth related actuivity and
```
avichalp: `s/actuvity/activity/`

senic_hub/backend/nuimo_setup.py
```diff
@@ -27,96 +85,230 @@ def discover_and_connect_controller(self, required_mac_address=None, timeout=Non
...
+
+        logger.debug("Stopping nuimo discovery")
         self._manager.stop_discovery()
-        self._connect_timeout_timer = threading.Timer(20, self.connect_timed_out)
+
+        self.start_connection(discovered_nuimo)
+
+    def start_connection(self, discovered_nuimo):
+        """
+        Blocking. Starts the nuimo discovery.
```
avichalp: >Starts the nuimo discovery.

Should it be start 'Starts the nuimo connection'? Guessing it by the name of the function.

senic_hub/backend/commands.py
rudolfs: +1

TheMeaningfulEngineer:
@rudolfs It is addressing a single thing which is a successful on boarding of a single Nuimo.
I want to merge it into a branch that will contain all the BT related fixes, not into master.
Pls check here for the overview: https://github.com/getsenic/senic-hub/issues/394

It is focused on a **single Nuimo** use case.

>  onboarding works great, tried multiple times, with different WiFis and different devices

That's good :)

> after the hub is rebooted everything is broken, same if I restart all of the services with supervisorctl restart all

This is being addressed here: https://github.com/getsenic/senic-hub/issues/408

> "add NUIMO" not working, always detects the same nuimo
> when removing a nuimo "add NUIMO" stops working at all

These two things are for multiple Nuimos use case. They need to be addressed 
for the MVP but not for the SD flashing.
Can you please add them as separate issues in the hub?

TheMeaningfulEngineer:
@avichalp 

> Restarting as a solution and expecting that discovery and connection works next time doesn't look like a robust solution.

Fair enough :)
It is the best one we have at the moment, and we need to try and break it.

> What is the fallback when max number of retries exceed?

There isn't one, we're concluding that there just isn't a Nuimo present.
Which might mean:

* an error on the Nuimo
* a firmware issue on the BT module
* a hardware issue related to the BT module
* a Nuimo really isn't present or is turned off 

> In which case the discovery succeeds but connect times out?

There are situations where you can discover a device but you
can't connect to it, they depend on the BT connection quality. 
I've attributed it to the underlying physics of wireless 
communication. Detecting a single message being broadcasted all the time
is a more probable action to occur in an environment of bad connectivity as 
opposed to a dedicated exchange of BT messages (needed for a connect).

avichalp:
@TheMeaningfulEngineer,
Thanks for explaining the discover v/s connect and fallback scenario. However I am still not convinced about the 1st point.

> It is the best one we have at the moment, and we need to try and break it.

Even if this is the best we have but it is too complicated to understand that is why it will be difficult to maintain in future. I understand there is already a pressing deadline but if we can give more time on it I would vote for spending more time.

By looking at the bug reports (issues), logs data on sentry and and some other data I created a hypothesis around this issue. Since I do not have the adequate hardware I am not able to test my hypothesis. However I would like to share it with a statutory warning that it could be BS but in case if it is helpful. 

I think something is up with this line here:
https://github.com/getsenic/senic-hub/pull/411/files#diff-fcecbd5461c7c1efc43b71ff1271ca7bL13
```
     def __init__(self, adapter_name):
        self._adapter_name = adapter_name
```
`adapter_name` is being passed to this function from here:
https://github.com/getsenic/senic-hub/blob/8123a2bbb72e5586a49a5e71d48939997f3e2df8/senic_hub/backend/views/nuimos.py#L62-L65

But this part of the code gets the adapter name from `senic_hub.ini` which has a fixed value of `hci0`.  But few days back we [observed](https://github.com/getsenic/senic-os/commit/3abed178b7d55a0e5ea1870efee9df6f3486ccf6) that `adapter_name` sometimes changes in new BT module (it is not always `hci0`) and this change is somewhat random and happens on restart.

I think the error happens when we have wrong `adapter_name` (may be `hci1`) and fixes itself when after restart it gets changed back to `hci0`. 

This hypothesis can be checked by logging the value of `adapter_name`. 
I will not be able to test this but would be interested to know if I am thinking in correct direction.
What do you think?

TheMeaningfulEngineer:
@avichalp 
> This hypothesis can be checked by logging the value of adapter_name.

Yes! 
Great observation. Will add that to the logging.

> Even if this is the best we have but it is too complicated to understand that is why it will be difficult to maintain in future. I understand there is already a pressing deadline but if we can give more time on it I would vote for spending more time.

Conceptually it's "restart the bt module after timout", and I agree that we should try and make the implementation of it as simple as possible. 
Even if the if the `adapter_name` proves to be the thing, we have no ground at the moment to trust the BT module and use that as a safe ground to simplify the higher level app stack.


Big like for this:
https://github.com/getsenic/senic-hub/blob/8123a2bbb72e5586a49a5e71d48939997f3e2df8/senic_hub/backend/views/nuimos.py#L62-L65
a huge plus for communicating clearly. Will adopt it :)

rudolfs:
@TheMeaningfulEngineer 
> These two things are for multiple Nuimos use case. They need to be addressed
for the MVP but not for the SD flashing.
> Can you please add them as separate issues in the hub?

Added an issue here: https://github.com/getsenic/senic-hub/issues/415

senic_hub/backend/nuimo_setup.py
TheMeaningfulEngineer: Beautifully phrased :)

TheMeaningfulEngineer:
@rudolfs 
> Could be a good workaround for now... 

For what problem?

rudolfs:
@avichalp @TheMeaningfulEngineer Just tried to dynamically get the BT interface name and supply it in the application layer (nuimo_app, backend api) and it didn't solve anything. I think the problem is in the GATT library.

senic_hub/backend/nuimo_setup.py
```diff
@@ -27,96 +91,229 @@ def discover_and_connect_controller(self, required_mac_address=None, timeout=Non
...
+        """
+        Blocking. Detects and connects to a nuimo, restarts BT on timeout.
+
+        Initiates the discovery process which blocks until the nuimo
+        is found and connected or it timed out.
+        """
+        logger.debug("Staring discovery timer [%ss]" % DISCOVERY_TIMEOUT)
+        self._discovery_timeout_timer = threading.Timer(DISCOVERY_TIMEOUT, self.discovery_timed_out)
+        self._discovery_timeout_timer.start()
+        self._start_discovery()
```
avichalp: Could `_start_onboarding` and `_start_discovery` be one one method?
Looks like `_start_discovery` is not being called from any where else besides `_start_onboarding`. 
Also, to me it looks like both methods are trying to achieve very similar things setting timer and starting the Glib loop for listening to callbacks.

What do you think?

TheMeaningfulEngineer: They could :)
My initial idea the split was to try and provide a higher level method in which you're not worrying about the distinctions between connect and discover, it's totally open to reimplementation at some point, but won't be going through that for this merge (it've tested with the PR changes and rebased new branches on them already)

TheMeaningfulEngineer:
Peeps. 
I'm merging this is so I don't end up in to many branches.
Am just making a PR for this https://github.com/getsenic/senic-hub/issues/407 that 
is based on this PR and retains the vanilla onboarding functionality.

PR #424, opened 2018-02-13
Reconnect to a nuimo after a long period of no connection

TheMeaningfulEngineer:
This was a method I initially removed in hope to approach it
somewhat safer.
The issue is that it's initiating loads of connects and not waiting
for them to fail. It's not known how this is handled in the underlying
libraries but the basic use case tests show it working.

The Nuimo successfully reconnected after 2 minutes of being off.

The conditionals in the method have been split to observe in
which way self.controller.is_connected() and self.connection_failed
relate to each other.


senic_hub/nuimo_app/__init__.py
```diff
@@ -263,6 +273,17 @@ def get_prev_component(self):
         else:
             return self.components[0]
 
+    def check_nuimo_connection(self):
+        while True:
+            time.sleep(NUIMO_RECONNECT_POLL_PERIOD)
+            if not self.controller.is_connected():
+                logger.debug("Not detecting a connected Nuimo")
+                if self.connection_failed:
```
avichalp: Question: 
Is there any case where nuimo is not connected `self.controller.is_connected()` is false and `self.connection_failed` is false too?

senic_hub/nuimo_app/__init__.py
```diff
@@ -128,6 +134,10 @@ def start(self, ipc_queue):
                             daemon=True)
         ipc_thread.start()
 
+        logger.debug("Started the connection check thread. Poll interval [%ss]" % NUIMO_RECONNECT_POLL_PERIOD)
```
avichalp: how about breaking the line after 79 chars like you did in the log below? I think it is more readable that way.

senic_hub/nuimo_app/__init__.py
TheMeaningfulEngineer: I don't know.
In the [original](https://github.com/getsenic/senic-hub/blob/af8069eaa803824d1e202676d66dde48add76a3c/senic_hub/nuimo_app/__init__.py#L305) (just some random commit before my changes) the seem to be checking for everything at once.

I've split it up to gather some data through usage and see if they always occur in parralel.

senic_hub/nuimo_app/__init__.py
TheMeaningfulEngineer: :+1:

senic_hub/nuimo_app/__init__.py
avichalp: Alright 👍

Issue #425, opened 2018-02-14
Add helper functions module


avichalp:
https://github.com/getsenic/senic-hub/blob/master/senic_hub/backend/views/appinfo.py#L28-L67

These functions here are generic and can be used at multiple places where we need the data about the os or the hardware etc. One example is we need the os version to append it to the log which we can send to sentry.

Would be good to create a module `senic_hub.backend.utils` where we can move these three functions to start with. Later we can put other utility functions here that can be used across `backend` module.

Thoughts?
CC: @rudolfs

rudolfs:
+1

Bonus points if you put it in a class and write some tests.

avichalp:
I can write tests around it. But did not understand why you would like to put them in a class?

rudolfs:
I guess I'm just used to Ruby, where everything is a class. Since all of these functions are extracting information about the system I'd put them in a class that deals with just this.

Do whatever is idiomatic Python and it'll be good.

avichalp:
initial attempt: https://github.com/getsenic/senic-hub/pull/426

PR #426, opened 2018-02-14
Move common functions to senic_hub.backend.utils

avichalp:
Move data related functions from `appinfo` to `utils` and added tests.

Connect #425


senic_hub/backend/views/appinfo.py
```diff
@@ -19,49 +15,7 @@
 @app_info.get()
 def get_app_info(request):
     return dict(
-        wifi=wifi(),
-        os_version=os_version(),
-        hardware_identifier=hardware_identifier(),
+        wifi=utils.read_wifi('/etc/NetworkManager/system-connections/bluenet'),
+        os_version=utils.read_os_version('/etc/os-release'),
+        hardware_identifier=utils.read_hardware_identifier('/proc/cpuinfo'),
```
rudolfs: Why do we want to pass the file paths? I think the function should hide that implementation detail.

avichalp: 1. To make it easier for testing.
2. file path is not implementation detail it is data.
3. Implementation detail should be hidden not data (So we can mock it with different data easily for same implementation).
4. Here implementation details are the logic that performs and `split`, `replace`, `read` .. etc.

Thoughts?

rudolfs: I don't see where the same function should ever be used with different paths. This means, that we always invoke each function with the same path. Why repeat yourself?

In this case the implementation assumes a certain file, so the path is an implementation detail. We can't pass the function any file and expect it to give us a hardware identifier.

You would just have to stub the call to `open()` with the specific path for each function in the tests. Is that hard to do in Python?

avichalp: >I don't see where the same function should ever be used with different paths. This means, that we always invoke each function with the same path. Why repeat yourself?

Got your point about not using the same function elsewhere. It make sense.

>You would just have to stub the call to open() with the specific path for each function in the tests. Is that hard to do in Python?

Not at all. It can be easily done by mocking.

What do you think about doing the `open` call outside these functions and passing only the string to these functions. Then each function would just implement the parsing logic for the string it received. I suggest this is because I would prefer having pure functions (if feasible) rather than mocking them for tests.

rudolfs: I would just convert the whole thing to a class, the public interface would be something like this:

HubMetadata#current_connected_wifi()
HubMetadata#os_version()
HubMetadata#hardware_identifier()

Then you can have private methods that help you do whatever your public methods need in a DRY way. As for the tests some mocking still would be required, but you only test the public interface.

avichalp: Okay, this make sense.
changed as you suggested.

rudolfs:
Looks great. Could you squash it into one commit and then merge into dev?

avichalp:
@rudolfs : Thanks for the review. Merging.

Issue #417, opened 2018-02-07
Clean up hardcoded IP


rudolfs:
In `senic_hub/backend/device_discovery.py`:
`self.app_name = "senic hub#192.168.1.12"  # TODO get real IP of the hub`

PR #429, opened 2018-02-19
Use hardware id to authenticate with Philips Hue

rudolfs:
As the IP address could change when the DHCP lease expires.

Connects #417


senic_hub/backend/device_discovery.py
```diff
@@ -214,7 +215,7 @@ class PhilipsHueBridgeApiClient:
     def __init__(self, ip_address, username=None):
         self.ip_address = ip_address
         self.bridge_url = "http://{}/api".format(self.ip_address)
-        self.app_name = "senic hub#192.168.1.12"  # TODO get real IP of the hub
+        self.app_name = "senic_hub#" + HubMetaData.hardware_identifier()
```
rudolfs: You can check how requests should look like here: https://developers.meethue.com/documentation/getting-started

Issue #403, opened 2018-01-29
Provide more context to sentry errors


kelada:
Can you provide more context to a stacktrace, more specifically:
- version of senic os
- version of Philips hue api (and if possible software version of the bridge)
- version of sonos api (and if possible software version sonos speaker)

This will allow us to spot issues around device versions changing outside of our control.

TheMeaningfulEngineer:
@avichalp This can all go into the `improve-logging`. I'm regularly rebasing on that for the BT issue.

avichalp:
These are interesting piece of information.
How about adding `senic-os` version in generic log formatters?
For eg: `2018-01-29 14:54:55,814 <os-version> INFO  [senic_hub.nuimo_app] gesture BATTERY LEVEL 10`

TheMeaningfulEngineer:
Not up for that solution, seems unnecessarily redundant to have it in every log line.

burakparim:
Having `senic-os` version on logs might be helpful I think, not now but in the future. Do you know if error rate vs battery level of a nuimo is measured? Eg., stations/ring stays idle half of the time below 15% battery... This is not hub related, but we can improve Nuimo user exp if we detect this sort of issues.

avichalp:
@burakparim @TheMeaningfulEngineer 
I also think that `senic-os` level can be helpful in future specially when we start to support multiple os builds. Lets discuss more about merits and demerits here.

>do you know if error rate vs battery level of a nuimo is measured?
This is currently not logged. But I agree that this is a interesting use case. As you pointed out it is not related to the hub and I also would I like to take a look into it separately. If you can create a feature request on a different Github issue we can discuss about it there in more detail there.

PR #430, opened 2018-02-20
Extra logs info

avichalp:
This PR aims to make additional info about philips Hue bridge available to the application. So this information can be used for logging purpose. 
In this PR I have added `phue_bridge_info` on `HubMetaData` class to get the info about the bridge when required.
https://github.com/getsenic/senic-hub/blob/4e71b536b287bc9670266278835f8bb3ccd4d31f/senic_hub/backend/hub_metadata.py#L58-L68

This information is saved in the `devices.json` file while setting up the Philips Hue:
https://github.com/getsenic/senic-hub/blob/4e71b536b287bc9670266278835f8bb3ccd4d31f/senic_hub/backend/views/setup_devices.py#L174-L178

Here is an example where having philips hue API and software version helps in error logging.
https://github.com/getsenic/senic-hub/blob/4e71b536b287bc9670266278835f8bb3ccd4d31f/senic_hub/backend/views/philips_hue.py#L69-L81

Connects #403 

In this PR I aim to add an API (`phue_bridge_info`) to get the required info. Future PRs related to this issue will add more logs (where required) and improving existing logs by adding more info about the devices using this API. And also, add a similar API for `sonos`.


senic_hub/backend/tests/data/devices.json
```diff
@@ -18,11 +18,17 @@
...
-	"5": {"name": " Light 5"},
-	"7": {"name": " Light 7"},
-	"4": {"name": " Light 4"},
-	"8": {"name": " Light 8"},
-	"6": {"name": " Light 6"}
+	      "5": {"name": " Light 5"},
+	      "7": {"name": " Light 7"},
+	      "4": {"name": " Light 4"},
+	      "8": {"name": " Light 8"},
+	      "6": {"name": " Light 6"}
```
rudolfs: Don't mix tabs with spaces.

senic_hub/backend/hub_metadata.py
```diff
@@ -15,7 +16,10 @@ def _read_from_file(cls, file_path):
             return ''
 
         with open(file_path) as data_file:
-            return data_file.readlines()
+            if file_path.endswith('.json'):
+                return json.load(data_file)
+            else:
+                return data_file.readlines()
```
rudolfs: Can it ever be a non-json file?

senic_hub/backend/views/philips_hue.py
```diff
@@ -10,6 +10,7 @@
 from colander import MappingSchema, SchemaNode, String, Int, Range, Length
 from cornice.validators import colander_body_validator
 from .api_descriptions import descriptions as desc
+from .. import hub_metadata
```
rudolfs: Why is this extra namespace needed? Is this something idiomatic to Python? Importing HubMetaData would be more concise and then using it like this:
`HubMetaData.os_version()`

senic_hub/backend/hub_metadata.py
avichalp: Yes, this the same function that reads different files like: os-release, devices.json etc.
If it finds that the json extension it marshals the `json` into `python dict` otherwise reads the file line by line.

senic_hub/backend/views/philips_hue.py
avichalp: Makes complete sense. 
We should use : 
`import <module or pacakge>` : for importing modules and packages.
`from module import <class or function>`: for Importing class/functions/objects etc.
Reference: [Google python style guide](https://google.github.io/styleguide/pyguide.html#Imports)

Thanks for catching.

senic_hub/backend/views/philips_hue.py
```diff
@@ -10,6 +10,7 @@
 from colander import MappingSchema, SchemaNode, String, Int, Range, Length
 from cornice.validators import colander_body_validator
 from .api_descriptions import descriptions as desc
+import senic_hub.backend.hub_metadata as hub_metadata
```
rudolfs: Why do we need the extra `hub_metadata` namespace and not just the class that you will be using? Is that something idiomatic to Python?

avichalp: Not related to `python`. But in general one advantage that we have by importing module and not class/functions directly is that in future it reduces the chance of name conflicts. 
For example:  when you have 2 namesapces that provide functions with same name it can conflict in a module where both of them are imported.

rudolfs: I see, so it's just a precaution.

kelada:
There are loads of constants that should be extracted to dedicated class and imported. I won't require it for this PR because it's a general refactoring, but it should be done.

PR #432, opened 2018-02-22
Nuimo reconnects by a self restart of the nuimo_pp

TheMeaningfulEngineer:
We need this for the internal beta to provide a
sub-optimal but usable solution.
It is possible that the implementation will change we
start doing actual changes to BlueZ (less hacks might be needed here).

The nuimo will reconnect eventually.
Seems the addition of sys.exit speed up things.

This is a hacky workaround that isn't scalable to multiple
nuimos and has reconnect times that feel too long
(my subjective feeling).

The issue that is being worked around is that the `connect`
sometimes just stop responding to the Nuimo even when it's in range.

This has been explored down to this call:

```
senic-hub bluetoothd[2019]: ../bluez-5.46/src/device.c:device_connect_le() Connection attempt to:
 F5:D4:41:9C:86:B0
```

(output from `journalctl --unit=bluetooth -f` with debug logs enabled)


senic_hub/nuimo_app/__init__.py
```diff
@@ -170,6 +176,7 @@ def start(self, ipc_queue):
         self.set_active_component()
         logger.info("Connecting to Nuimo controller %s", self.controller.mac_address)
         self.controller.connect()
+
```
avichalp: Good one 👍

